PR #31, opened 2024-12-21
Added note about cleaning workspace when cached files change

DingoOz:
docu: Added tip about cleaning the workspace if there failing colcon build due to a GH update or indirect dependency change.


RandomSpaceship (approved):
Good overall.
Just converted quotes -> code blocks, reworded slightly, and added recommendations on when to run a clean,

PR #39, opened 2024-12-24
Networking library implementation

RandomSpaceship:
RAII wrapper around sockets for our use-cases. Will allow cleaning up `sensors` node.


DingoOz:
@RandomSpaceship code looks great. Solid RAII and Rule of 3/5 applied with fd_wrapper and using C++20.  Names appear to all conform to the software standards. Just one comment to consider re the documentation for class Client.

DingoOz (approved):

PR #41, opened 2024-12-26
Xbox controller node improvement

DingoOz:
This PR updates the Xbox controller node code that existed in main before the adoption/enforcement of the software standards.

This update to the python file for this node does:
- updates the node name to fit with the software standards
- adds a second deadman's switch which when used doubles the speeds (turbo)
- adds the ability to control the scale of movements via parameters
- adds the ability to output TwistStamped instead of Twist via a parameter (defaults to Twist)

New parameters are:
translation_scale (double): Base scaling for linear motion [m/s] (default: 0.25)
rotation_scale (double): Base scaling for angular motion [rad/s] (default: 0.50)
high_speed_multiplier (double): Multiplier for high-speed mode (default: 2.0)
deadman_threshold (double): Threshold for dead man's switch activation (default: -0.95)
use_stamped_msg (bool): Whether to publish TwistStamped instead of Twist (default: False)

### DIFF
--- a/software/ros_ws/src/input_devices/input_devices/xbox_controller.py
+++ b/software/ros_ws/src/input_devices/input_devices/xbox_controller.py
@@ -11,8 +11,8 @@ class XboxController(Node):
     A ROS2 node that processes Xbox controller inputs and publishes velocity commands.
 
     This node subscribes to joystick inputs and converts them into velocity commands
-    for robot control. It includes safety features like a dead man's switch and
-    configurable scaling factors.
+    for robot control. It includes two dead man's switches for different speed modes
+    and configurable scaling factors.
 
     Publishers:
         input_devices/cmd_vel (geometry_msgs/Twist): Robot velocity commands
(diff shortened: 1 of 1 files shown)


software/ros_ws/src/input_devices/input_devices/xbox_controller.py
```diff
@@ -1,67 +1,222 @@
...
+        )
+
+        self.declare_parameter(
+            "use_stamped_msg",
+            False,
+            ParameterDescriptor(
+                description="Whether to publish TwistStamped instead of Twist messages",
+                read_only=True,
+            ),
+        )
```
RandomSpaceship, line 124: Although these parameters are declared as run-time modifiable (`read_only=False`), you aren't polling for changes or using [callbacks](https://docs.ros.org/en/jazzy/Tutorials/Intermediate/Monitoring-For-Parameter-Changes-Python.html) to monitor it, so any changes after initialisation won't properly be applied.

RandomSpaceship (approved):
Looks pretty good! Just the one note on not monitoring for parameter changes, but I'm happy for this to be merged as-is as long as you raise an issue so we have it tracked that we need to deal with it.

software/ros_ws/src/input_devices/input_devices/xbox_controller.py
DingoOz: https://github.com/ROAR-QUTRC/perseus-v2/actions/runs/12503292461

DingoOz:
You are right, we need to have runtime changes handled proper. I didn't want to leave this ready to bite us later on so I've updated for this PR.

PR #52, opened 2025-01-05
docs: Fix issue with getting started

RandomSpaceship:
The cwd needs to be the repo when you run the getting-started script.


DingoOz (approved):
Looks good.

PR #58, opened 2025-01-06
Device-specific setup, general bugfixes

RandomSpaceship:
Features:
- Home Manager set up to allow for standardised environment across devices
- `scripts/machine-setup.sh` script written to automatically apply first-time Home Manager setup
- `direnv` setup message now colourised (orange) to make it stand out
- CAN bus configuration now managed with Home Manager
- Added `apps.perseus` (currently runs Perseus ROS bringup, but may be expanded in the future) - aliased to `apps.default` to keep previous behaviour

Also fixed a whole bunch of bugs/minor issues encountered during the install and testing.

Bugfixes:
- Scripts for which it matters now check if they're run as root and exit with an error message
- Several scripts updated to be brittle (`set -euo pipefail`)
- Nix setup now adds Cachix repos to `extra-trusted-substituters` rather than adding the user to `trusted-users`, fixing a potential security vulnerability
- CI/CD now also runs on `merge_group`
- Running `bash` in the dev environment now correctly uses `bashInteractive`
- `treefmt.toml` is now no longer architecture-specific


.envrc
DingoOz, line 1: nice one, this is important to have in colour.

DingoOz (commented):
thank you :)

DingoOz (approved):
All looks good - I am not accross most of this but I can see the thoiught and hard work in this.Thanks.

PR #69, opened 2025-01-10
Autonomy: CMakeLists.txt and launch file aligned with Project

DingoOz:
The Autonomy code existed before the Perseus Project Software Standards. This is a fix to make the CMakeLists.txt and launch file more aligned with project formatting and approach and housing keeping on package.xml and README.md.

This update of legacy code is to support a separate future feature for slam toolbox based mapping.

### DIFF
--- a/software/ros_ws/src/autonomy/CMakeLists.txt
+++ b/software/ros_ws/src/autonomy/CMakeLists.txt
@@ -1,5 +1,10 @@
-cmake_minimum_required(VERSION 3.8)
-project(autonomy)
+# PROJECT SETUP
+cmake_minimum_required(VERSION 3.23)
+
+project(
+  autonomy
+  VERSION 0.0.1
+  LANGUAGES CXX)
 
 # credit https://www.kitware.com/cmake-and-the-default-build-type/
 set(default_build_type "Debug")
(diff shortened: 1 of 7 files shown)


software/ros_ws/src/autonomy/CMakeLists.txt
```diff
@@ -24,17 +34,45 @@ if(CMAKE_BUILD_TYPE MATCHES "Rel.*")
...
+    rclcpp
+    slam_toolbox
+    nav2_bringup
+    sensor_msgs
+    geometry_msgs
+    tf2
+    tf2_ros)
+
+# packages which must be present but aren't linked against
+set(THIS_PACKAGE_BUILD_DEPENDS backward_ros ament_cmake)
```
RandomSpaceship, line 52: Dependency list is incorrect

software/ros_ws/src/autonomy/CMakeLists.txt
```diff
@@ -24,17 +34,45 @@ if(CMAKE_BUILD_TYPE MATCHES "Rel.*")
...
+  find_package(${Dependency} REQUIRED)
+endforeach()
+
+# Set up testing configuration
 if(BUILD_TESTING)
   find_package(ament_lint_auto REQUIRED)
   set(ament_cmake_copyright_FOUND TRUE)
   set(ament_cmake_cpplint_FOUND TRUE)
   ament_lint_auto_find_test_dependencies()
 endif()
```
RandomSpaceship, line 68: Delete this

software/ros_ws/src/autonomy/launch/mapping_using_slam_toolbox.launch.py
RandomSpaceship, line 1: Should probably split out autonomy package launch to be its own thing so we're not duplicating code with the main launch file (ie, import a launch file), but that can be fixed later (I'll probably be reorganizing the launch files as part of the sim bringup, so don't worry about it)

RandomSpaceship (changes requested):

DingoOz:
Thanks for the review. Items updated, ready for final review.

software/ros_ws/src/autonomy/CMakeLists.txt
```diff
@@ -24,17 +34,30 @@ if(CMAKE_BUILD_TYPE MATCHES "Rel.*")
...
-endif()
+# Set the C++ standard
+set(CMAKE_CXX_STANDARD 20)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+
+# INCLUDE DEPENDENCIES public dependencies (which should propagate)
+set(THIS_PACKAGE_INCLUDE_DEPENDS slam_toolbox geometry_msgs tf2_ros)
+
+# packages which must be present but aren't linked against
+set(THIS_PACKAGE_BUILD_DEPENDS backward_ros ament_cmake)
```
RandomSpaceship, line 45: because you're not building a target (and therefore not linking against anything), you can strip out everything but `ament_cmake` here.

RandomSpaceship (changes requested):
Aside from `ament_cmake`, which provides the ament packaging macros, you only need to run `find_package` in the `CMakeLists.txt` file when you're _building_ a C/C++ target (and therefore need to link libraries against it). Actually pulling the libraries in is specified in the `package.xml`.

software/ros_ws/src/autonomy/CMakeLists.txt
DingoOz: ahhh yes, good point.

DingoOz:
Thanks, updates made based on review feedback, ready for review again

RandomSpaceship (approved):

RandomSpaceship:
Looks good!

PR #74, opened 2025-01-12
Documentation(software): Section for simple-networking library

DingoOz:
added a documentation section for simple-networking and added a tip under shared libraries for running nix-package.sh


RandomSpaceship (approved):
Looks good. It's missing a fair bit of nuance, but it'll be quicker for me to write that documentation myself when I'm back than to explain it and have you write it. Still significantly better than having nothing at all.

PR #79, opened 2025-01-16
Feat/handheld mapping

DingoOz:
Merged three branches into one. PR covers:
*adding rqt and tf2_tools to the flake.nix
*adding slam_toolbox and launch files for autonomy
* updating autonomy ready for near future feature addition of using robot_localization 
* fixed IMU data's frame so it is separate to laser_scan data
* correcting m2m2 lidar so that is produces 4096 interpolate points per scan


software/ros_ws/src/perseus_sensors/src/m2m2_lidar/m2m2_lidar.cpp
```diff
@@ -670,6 +554,183 @@ void M2M2Lidar::_readSensorData()
...
+    vector<tuple<float, float, bool>> points;
+    points.reserve(input_points.size());
+
+    // Normalize angles to [0, 2π]
+    for (const auto& point : input_points)
+    {
+        float angle = get<0>(point);
+        // Normalize angle to [0, 2π]
+        while (angle < 0) angle += 2 * M_PI;
+        while (angle >= 2 * M_PI) angle -= 2 * M_PI;
```
RandomSpaceship, line 636:
```suggestion
        const auto tau = 2 * std::numbers::pi;
        angle -= std::trunc(angle / tau) * tau;
        if (angle < 0) angle += tau;
```
While loops to process any kind of number are a bad idea, and especially so for floats.

If you use `std::numbers::pi` (yay C++20!) instead of `M_PI` from `<cmath>` you'll need to `#include <numbers>` too

software/ros_ws/src/perseus_sensors/src/m2m2_lidar/m2m2_lidar.cpp
```diff
@@ -670,6 +554,183 @@ void M2M2Lidar::_readSensorData()
...
+    for (size_t i = 0; i < INTERPOLATED_POINTS; ++i)
+    {
+        float targetAngle = i * angleIncrement;
+
+        // Find points that bracket our target angle
+        size_t idx = 0;
+        while (idx < points.size() && get<0>(points[idx]) < targetAngle)
+        {
+            idx++;
+        }
```
RandomSpaceship, line 661: Again, a while loop is a bad idea. You've already sorted your points, and `size(interpolated points) >> size(source points)`, so you can store `idx` outside the loop (though you should probably rename it to something like `sourcePointIdx`), and just do a check to see if you need to increment it once at the start of each loop.

RandomSpaceship (commented):
Overall good work, although there are a few issues.
- Some launch files are made obselete by work on `gazebo-sim`, and should probably be refactored to better split responsibilities
- I haven't reviewed the parameter files due to a lack of time

docs/source/architecture/software.md
```diff
@@ -103,6 +107,17 @@ However, this on its own is not particularly useful - which is where implementat
 The `hi-can-raw` library implements {class}`hi_can::FilteredCanInterface` using the Linux SocketCAN [`RAW_CAN`](https://docs.kernel.org/networking/can.html#raw-protocol-sockets-with-can-filters-sock-raw) interface, and is what most code uses to interface with the CAN bus.
 The `hi-can-net` library is currently unused, but may become an implementation of {class}`hi_can::CanInterface` which forwards all traffic over a network connection.
 
+#### Simple-networking
+
+The simple-networking library provides a modern C++ implementation for handling network socket communications, with a primary focus on client-side operations.
+
+It offers an object-oriented wrapper around traditional POSIX socket operations, supporting both TCP and UDP protocols. The library implements RAII principles through its Client class, which manages socket creation, configuration, connection and cleanup while providing exception-based error handling for robust failure management.
```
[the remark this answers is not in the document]
DingoOz: Updated the .md with networking::client

software/ros_ws/src/autonomy/launch/handheld_mapping.launch.py
```diff
@@ -0,0 +1,229 @@
...
+            executable="static_transform_publisher",
+            name="imu_tf_publisher",
+            arguments=["0", "0", "0.1", "0", "0", "0", "base_footprint", "laser_frame"],
+        ),
+        Node(
+            package="tf2_ros",
+            executable="static_transform_publisher",
+            name="imu_tf_publisher",
+            arguments=["0", "0", "0.4", "0", "0", "0", "base_footprint", "imu_frame"],
+        ),
```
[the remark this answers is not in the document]
DingoOz: This code is for handheld mapping and hence just publishes static transforms instead of relying on Perseus' tf. The tf joint names are aligned with Slam Toolbox's default names.

software/ros_ws/src/perseus_description/urdf/perseus.urdf.xacro
```diff
@@ -8,6 +8,18 @@
...
+    <joint name="base_footprint_joint" type="fixed">
+      <parent link="base_link"/>
+      <child link="base_footprint"/>
+      <origin xyz="0 0 0" rpy="0 0 0"/>
+    </joint>
+
+    <link name="base_footprint">
+    </link>
+
+
```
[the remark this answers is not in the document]
DingoOz: Slam toolbox does not show a lot of internal information when it runs. Its documentation relies on several example launch files. The default tf tree structure for slam toolbox includes "base_footprint" instead of "base_link" so that is why this was added. Since there is no actual Perseus to test on currently, attempting to make as few assumptions with slam toolbox as possible until we can test that a remapping or the slamtoolbox yaml will correctly consumes base_link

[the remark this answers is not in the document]
DingoOz: Since we don't have actual Perseus to test on yet, I have reverted to base_link and expect the params to be applied to slam_toolbox.

software/ros_ws/src/autonomy/launch/mapping_using_slam_toolbox.launch.py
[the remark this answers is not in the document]
DingoOz: I think I know what is needed here. Change the launch file to call the perseus bringup launch file.

DingoOz:
Updated all items raised, cleaned up commits, should be ready for final review thanks.

software/ros_ws/src/autonomy/launch/handheld_mapping.launch.py
RandomSpaceship, line 1: I think we should stick to a single way of writing launch files using array concatenation in the constructor, rather than using both that and `add_action` without a good reason (I'd be fine with either, but at the moment we have more of the former than the latter). If you think it's a bad idea, I'm happy to discuss that though.

software/ros_ws/src/autonomy/launch/mapping_using_slam_toolbox.launch.py
RandomSpaceship, line 1: Looks good. If you want, I made a `nixgl` command/script which you can use instead of `nix run...` and environment variables (ie, `nixgl rviz2` instead of `NIXPKGS_ALLOW_UNFREE... nix run --impure ... -- rviz2`), which I think is slightly cleaner. Aside from that, this all looks great.

software/ros_ws/src/perseus_sensors/src/m2m2_lidar/m2m2_lidar.cpp
```diff
@@ -548,6 +555,210 @@ void M2M2Lidar::_readSensorData()
...
+        // Handle edge cases
+        if (sourcePointIndex == 0)
+        {
+            // Interpolate between last and first point
+            auto p1 = points.back();
+            auto p2 = points.front();
+
+            // Adjust angle for wrap-around
+            float a1 = get<0>(p1);
+            float a2 = get<0>(p2) + 2 * M_PI;
```
RandomSpaceship, line 700:
```suggestion
            float a2 = get<0>(p2) + 2 * std::numbers::pi;
```
You'll need to `#include <numbers>` though.

software/ros_ws/src/perseus_sensors/src/m2m2_lidar/m2m2_lidar.cpp
```diff
@@ -548,6 +555,210 @@ void M2M2Lidar::_readSensorData()
...
+            }
+
+            if (!get<2>(p2))
+            {
+                interpolated.emplace_back(targetAngle, get<1>(p1), true);
+                continue;
+            }
+
+            float t = (targetAngle - a1) / (a2 - a1);
+            float dist = get<1>(p1) + t * (get<1>(p2) - get<1>(p1));
```
RandomSpaceship, line 721: You should probably use [`std::lerp`](https://en.cppreference.com/w/cpp/numeric/lerp) instead

software/ros_ws/src/perseus_sensors/src/m2m2_lidar/m2m2_lidar.cpp
```diff
@@ -548,6 +555,210 @@ void M2M2Lidar::_readSensorData()
...
+            }
+
+            if (!get<2>(p2))
+            {
+                interpolated.emplace_back(targetAngle, get<1>(p1), true);
+                continue;
+            }
+
+            float t = (targetAngle - get<0>(p1)) / (get<0>(p2) - get<0>(p1));
+            float dist = get<1>(p1) + t * (get<1>(p2) - get<1>(p1));
```
RandomSpaceship, line 754: same thing with `std::lerp`

RandomSpaceship (changes requested):
Really good work, thank you! Just a few minor issues to fix, and it should be good to go after that.

docs/source/architecture/software.md
```diff
@@ -115,7 +115,7 @@ The `hi-can-net` library is currently unused, but may become an implementation o
 
 The simple-networking library provides a modern C++ implementation for handling network socket communications, with a primary focus on client-side operations.
 
-It offers an object-oriented wrapper around traditional POSIX socket operations, supporting both TCP and UDP protocols. The library implements RAII principles through its Client class, which manages socket creation, configuration, connection and cleanup while providing exception-based error handling for robust failure management.
+It offers an object-oriented wrapper around traditional POSIX socket operations, supporting both TCP and UDP protocols. The library implements RAII principles through its 'networking::Client' class, which manages socket creation, configuration, connection and cleanup while providing exception-based error handling for robust failure management.
```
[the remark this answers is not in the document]
RandomSpaceship: `nix build` always creates a `result` folder, or occasionally a file depending on what exactly you're doing - the docs should be under `result/html`. It's detailed [here](https://roar-qutrc.github.io/home/documentation-development.html#serving-the-built-docs).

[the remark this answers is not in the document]
DingoOz: Wonderful, thanks!

software/ros_ws/src/autonomy/launch/mapping_using_slam_toolbox.launch.py
DingoOz: If it is ok to leave this as is now and I'll start to use the nixgl script going forward.

RandomSpaceship: That's fine.

DingoOz:
@RandomSpaceship thanks for the review, updates made. Should be ready for review again please.

RandomSpaceship (approved):
Great work! LGTM.

RandomSpaceship (approved):

DingoOz:
@RandomSpaceship. Sorry about this, this PR had a five or so breaking merges which I had to resolve. Do you want to cast your eye over it again please?

RandomSpaceship (changes requested):
Just a few minor changes for some small issues, then good to go! Nice work.

DingoOz:
@RandomSpaceship All cleaned up and ready to go (hopefully)

software/ros_ws/src/autonomy/package.xml
RandomSpaceship, line 1: `exec_depend` actually is the right place for those dependencies since they're only used in Python files, and therefore only loaded at runtime. You should only need `depend` or `build_depend` if you need the packages from C++ code or CMake files. Doesn't matter though.

RandomSpaceship (approved):
Great work!

PR #84, opened 2025-01-18
documentation: Added a warning to stage ROS2 packages pre nix build

DingoOz:
Just added a warning to stage. I feel like there is a much large piece re new software packages etc so no point trying to detail this issue ahead of a more full treatment of the broader subject.


RandomSpaceship:
LGTM. Need to overhaul/write a lot of Nix usage docs at some point though, you're right.

RandomSpaceship (approved):

Issue #42, opened 2024-12-26
Demonstrate Gazebo Harmonic simulation to support autonomy


DingoOz:
Autonomy will require copious amounts of testing, some of which might benefit from simulation.

Gazebo Fortress is the preferred simulation for ROS2 Jazzy, and it is worth investigating the feasibility of implementing it with Perseus.

IssacSim requires nVidia hardware and can be considered in parallel.

RandomSpaceship:
I love the idea of running simulations with Gazebo, but I'm not super enthused about Isaac Sim and the like. If it ties us to specific hardware, it just makes it so much harder to test, as well as making it a pain for everyone who might not have that hardware to develop software.

Definitely needs investigation though, maybe Isaac Sim is good enough to justify the trouble.

DingoOz:
Yes, I agree focus should be on Gazebo only at first.

Hopefully a Gazebo implementation done smartly doesn't block other simulations laters being integrated

DingoOz:
Current blocker: ![Image](https://github.com/user-attachments/assets/82bc6e92-6d78-458e-b642-f16e608e5eba)

RandomSpaceship:
Simulation successfully demonstrated on branch `gazebo-sim`. Not quite ready for merging yet due to a last couple of minor issues (rover doesn't want to turn), but almost there.

PR #86, opened 2025-01-20
feat: Implement Gazebo simulation

RandomSpaceship:
Cleaning up and finishing the work done by @DingoOz to get a simulation running in Gazebo.

To test:
- Enter the simulation shell (`nix develop .#simulation`)
- Build and source the workspace
- Run `ros2 launch perseus_simulation perseus_sim.launch.py`
- In another shell (doesn't have to be the sim shell), run `ros2 run teleop_twist_keyboard teleop_twist_keyboard --ros-args -p stamped:=true` to test teleoperation

Known bugs:
- Turning currently doesn't work. The wheels spin correctly, but something about the physics isn't set up right, and it only moves forward/back.

Related: #42


DingoOz (approved):
Excellent work wrangling all of the different components here. Lots of of good stuff across all the files. I like the launch template, very handy for future as well.

I tested the gz-sim. Works as described. I was able to achieve turning but only with extremely high wheel speeds and then it massively under-steered, but it did "drift" and turn so concur that something in the physics sim needs to be tweaked as part of tuning but the code is working.

Well done!

PR #92, opened 2025-01-23
Bugfix/gazebo sim physics

DingoOz:
This PR changes a few wheel physics paramaters to allow for turning.
Also added a basic readme and a lunar world that is automatically downloaded from gz fuel.

### DIFF
--- a/software/ros_ws/src/perseus_simulation/README.md
+++ b/software/ros_ws/src/perseus_simulation/README.md
@@ -1,8 +1,16 @@
 # Running the Gazebo Simulation
 
-To run the most basic version of the simulation you will need to run the following in two separate termainals:
+This readme file explains how to launch the Gazebo Simulation environment and control the rover.
 
-## Terminal 1
+There is a separate nix devshell for simulation which is accessed via:
+
+```
+nix develop .#simulation
+```
+
+To run the most basic version of the simulation you will need to run the following in **two separate terminals**.
+
+## Terminal 1: Launch Simulation Environment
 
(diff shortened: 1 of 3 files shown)


software/ros_ws/src/perseus_simulation/README.md
```diff
@@ -0,0 +1,28 @@
+# Running the Gazebo Simulation
+
+To run the most basic version of the simulation you will need to run the following in two separate termainals:
+
+## Terminal 1
+
+To create the simulation:
+
```
RandomSpaceship, line 8: You should note that this needs to be done in a `nix develop .#simulation` shell

software/ros_ws/src/perseus_simulation/README.md
```diff
@@ -0,0 +1,28 @@
...
+```
+ros2 launch perseus_simulation perseus_sim.launch.py
+```
+
+## Terminal 2
+
+To send control commands using the keyboard:
+
+```
+ros2 run teleop_twist_keyboard teleop_twist_keyboard --ros-args -p stamped:=true
```
RandomSpaceship, line 18: This one's fine though, only applies to the sim

RandomSpaceship (approved):
Commit 6cd73fcd71c2baa577900002cd05b372e7ca5d25 has typos in it that you might want to fix before merging (not a big deal), and you should just note that the simulation needs to be run from a `simulation` shell in the README. It'll get added to the web docs eventually though, so I'm not super fussed.

PR #96, opened 2025-01-30
Implement LIDAR in Gazebo

RandomSpaceship:
Added a LIDAR on top of the Perseus model in Gazebo, in addition to an IMU (bridged to `/scan` and `/imu` respectively). Fixed some bugs I encountered along the way, including a proper fix for the wheel friction coefficients, allowing it to turn. Also added the empty world back in for easier testing.


UltraFishy (approved):

DingoOz (approved):
Nice work on the lateral friction.

PR #97, opened 2025-01-31
Update README.md

RandomSpaceship:
Update docs to reference the `nixgl` command I made


DingoOz (approved):
Good solution to the issue and well documented.

Issue #56, opened 2025-01-06
Investigate why CI/CD didn't run for #52


RandomSpaceship:
PR #52 had checks permanently listed as "pending". The initial push seemed to have the checks run against it, but the formatting commit that got pushed after it didn't - which is strange, because the format runs _before_ any of the other jobs. Perhaps it needs a PAT or deploy key so that the workflow re-runs after the format commit gets pushed*? I'd assume that GitHub associates the actions run with the commit that triggered it, not the commit that the checkout action pulls down, since that's the only thing that really makes sense to cause it.

*In this case concurrency should be set to cancel in progress.

RandomSpaceship:
Maybe helpful? https://stackoverflow.com/questions/72596035/how-to-associate-status-of-a-github-action-workflow-with-a-commit-or-pr

RandomSpaceship:
It looks like my initial hypothesis was correct, and the format commit doesn't get the status associated with it.

RandomSpaceship:
Same goes for any actions-generated commit by the looks of it, the run gets associated with the commit which triggered the run, but the branch protection rules require the _latest_ commit to have passing CI/CD, which isn't the same thing if a commit gets pushed from the workflow.

PR #107, opened 2025-02-04
Added the fonts required by zsh on Perseus' Orin

DingoOz:
- Orin shell is zsh so having fonts available for install will allow use by anyone even if the Internet is not available.


RandomSpaceship (approved):
Looks good, great idea.

RandomSpaceship:
Ah, looks like you need to push a change to the branch so that the actions will run again. It's #56 in action. I'd say just squash the Actions-generated `treefmt.toml` update into your last commit and force-push that.

PR #110, opened 2025-02-05
Bugfix - perseus_sensors (m2m2_lidar): bug fix that the version appended

DingoOz:
- Cmake was being asked to append the version number to the node name.

This branch ought to have been a bugfix in hindsight.


RandomSpaceship (approved):

RandomSpaceship:
Thanks for fixing that.

PR #112, opened 2025-02-05
perseus_simulation: changed the controller launch sequence

DingoOz:
- possible race condition on launch causing the timeout error, reverted the launch controller to sequence.
- also updated the frame for the lidar in the xacro


RandomSpaceship (approved):
Nice. Hopefully this fixes it once and for all!

PR #119, opened 2025-02-06
fix: Use correct axes for xbox deadman switch(es)

RandomSpaceship:
Needed to be axes 4 and 5, not 5 and 6. Now turbo mode works properly


OMEN44 (approved):

DingoOz:
sorry about that

RandomSpaceship:
Off by one comes for us all.

PR #122, opened 2025-02-12
docs: Updated teleop page with current network details.

DingoOz:


docs/source/home/teleoperation.md
```diff
@@ -89,6 +88,16 @@ Access Perseus via SSH from the laptop:
...
 ```
 
+### Manually confirm the date/time
+
+The Perseus compute modules (Orin, Pi5 etc) require appropriate date/time information. As they do not have a realtime backup clock source it is prudent to manually set the correct date/time after start up.
+
+If the Unifi UX has access to the Internet then it is possible to sync the time with
+
+```console
+sudo ntpdate pool.ntp.org
```
RandomSpaceship, line 98: `ntpdate` is [now considered deprecated](https://ubuntu.com/server/docs/about-time-synchronisation), and not installed by default. We should install `chrony` and just use that instead.

RandomSpaceship (changes requested):
Network instructions look good. Would just recommend changing the instructions to use `chrony` for setting the time, aside from that, it's good.

DingoOz:
updated with chrony :)

docs/source/home/teleoperation.md
```diff
@@ -89,6 +88,16 @@ Access Perseus via SSH from the laptop:
...
 ```
 
+### Manually confirm the date/time
+
+The Perseus compute modules (Orin, Pi5 etc) require appropriate date/time information. As they do not have a realtime backup clock source it is prudent to manually set the correct date/time after start up.
+
+If the Unifi UX has access to the Internet then it is possible to sync the time with
+
+```console
+sudo chronyc makestep
```
RandomSpaceship, line 98: Apparently you need the `-a` flag to get around 501 errors

RandomSpaceship: Or you could use `chronyd -q`

RandomSpaceship (approved):

PR #123, opened 2025-02-12
Bugfix/m2m2 lidar

DingoOz:
The m2m2 lidar driver had a startup bug causing timeout when it wasn't handling return codes properly. Fixed and additional logging too.


RandomSpaceship (approved):
Don't see any issues, though I also haven't tested it

PR #126, opened 2025-02-16
Feat/arm teleop direct script

DingoOz:
This is a stand alone program for direct control of a SO-100 Arm. 

Runs in the terminal, uses ncurses.


software/arm-teleop-direct/CMakeLists.txt
RandomSpaceship, line 1: Please use the CMake template snippets. You'll need, in order, 01, 15, 20, 22 (not 23, it's an executable so no public linking), and 71 (non-ROS). Just copy and paste them into a file in that order. The main thing is 01, since that sets up the compile rules as required by the software standards.

PR #137, opened 2025-02-23
perseus_sensors(m2m2_lidar): Added param for reading the imu or not

DingoOz:
Added a lot of debug info and made the Imu optional via a param - to give at least the option to avoid the imu read failure message.

IMU error not yet fixed by this PR but at least the laser scan data is available.


RandomSpaceship (approved):
LGTM!

PR #138, opened 2025-02-25
Resolved blocking issue for hi_can integration

GR3Y-SCALE:
Platformio was previously unable to build the esp32 firmware when both espidf and arduino frameworks were used, the issue was due to one of the modules included by espidf, esp__insights.
esp__insights required an ssl certificate to build into a c-object file, of which the ssl-certs location had to be specified in the platformio.ini file. 
`managed_components/espressif__esp_insights/server_certs/https_server.crt` is the location of the ssl certificate, managed components is a directory that is created when the project is built, so all the required certificates are included, the location of them needs to be specified.

esp__insights isn't used by hi_can and can be removed in the future, along with any other modules that aren't required, but for now it isn't included in the binary so it can be ignored for now, a later date can be arranged for these unused `managed_components` to be removed.


RandomSpaceship (approved):

BeHappyDog (approved):
Changes look good. Approving this.

PR #140, opened 2025-03-01
Update CI/CD to use ARM runner

RandomSpaceship:
Also fixing a few bugs while I'm at it, like docs failing to automatically deploy.


DingoOz (approved):
All looks good.

RandomSpaceship:
Added a `CODEOWNERS` file while I thought of it, set up automatic Nix packaging generation, fixed #56 (format + re-package commits now use a deploy key instead of the standard token, thus allowing Actions re-triggering), and added automatic skipping of relevant CI jobs on auto-generated commits.

BeHappyDog (approved):

PR #150, opened 2025-03-10
docs: Clarify minor points

RandomSpaceship:
Boolean usage and constant definition placement


DingoOz (approved):
Thanks for added the detail

PR #173, opened 2025-03-24
Add twist_mux node

RandomSpaceship:
Twist mux takes in, in order of increasing priority:
- cmd_vel
- key_vel
- joy_vel
And outputs the highest-priority option to cmd_vel_out

Also increased the diff drive controller timeout to allow for clock desync tolerance.


DingoOz (approved):
All looks good, good idea.

PR #174, opened 2025-03-27
Feat/mapping v3

UltraFishy:


software/ros_ws/src/autonomy/launch/cmd_vel_mux.launch.py
RandomSpaceship, line 1: this should probably be replaced with proper config of twist_mux in the main launch file, but this is fine for now

RandomSpaceship (approved):

DingoOz (approved):
send to Prod

PR #187, opened 2025-04-25
V0R1 E&C bucket PCB, KiCAD libraries

RandomSpaceship:
Exactly what it says in the title. Also some shared KiCAD libraries. This project will likely serve as a template for all future PCBs.


GR3Y-SCALE (approved):
Looks good.

PR #192, opened 2025-04-28
Feat/c1 lidar

DingoOz:
Adds the third party package to ros for rplidar. 

The existing rplidar-ros package in lop-sided-98's overlay is out of date (2.1.0) and this brings it up to latest version (2.1.5).


RandomSpaceship:
Good call, but the way you've implemented it currently is not the way to do it. To fix issues with packages that _already exist_ in the overlay, you should apply the fix in `ros_ws/patches.nix`. `third-party-packages` is **only** for software that need to be packaged _from scratch_. In this case, you can get away with overriding only the `version` and `src` fields (the former being to `` and the latter to what you currently have) of the pre-existing package.

Taking an example currently in the file (in this case, used to build fields2cover):
```
        tinyxml-2 = final.tinyxml-2.overrideAttrs ({
          src = final.fetchFromGitHub {
            owner = "leethomason";
            repo = "tinyxml2";
            rev = "c2d30872e20621955ca7feb9168bad996d591a19";
            sha256 = "sha256-Gn4d6v7p60XRam2wclaSFAiAxmNgKAKPxRCEmcMtJIE=";
          };
        });
```
You would just need to additionally override the `version` attribute to `2.1.5` (or `2.1.4` going off the `package.xml` file).

Finally, the root `flake.xml` isn't really the place to put the dependency. It's easy to do so for testing, but things shouldn't go there without a very good reason, and in this case, it can (and should) become a (runtime) dependency of `perseus_sensors`.

DingoOz:
Thanks on the feedback, yes I can see the better approach. I have implemented the preferred approach in a separate branch. Will shut this one down

PR #194, opened 2025-05-04
Feat/rplidar ros update

DingoOz:
patched the rosPackages to use ros2 version of rplidar-ros


software/ros_ws/patches.nix
```diff
@@ -18,6 +18,17 @@ let
           '';
       }
     );
+
+    rplidar-ros = rosPrev.rplidar-ros.overrideAttrs ({
+      version = "2.0.0"; # Version based on the ROS2 branch
```
RandomSpaceship, line 23: Should this not be what's in the `package.xml` (2.1.4)? That's what `ros2nix` uses for generating its version field.

flake.nix
```diff
@@ -156,6 +156,7 @@
             rqt-reconfigure
             rqt-common-plugins
             rmw-cyclonedds-cpp
+            rplidar-ros
```
RandomSpaceship, line 159: Can this be moved to a dependency of `perseus_sensors`?

RandomSpaceship (commented):
Looks good, big thing is just it being in `flake.nix`.

software/ros_ws/patches.nix
DingoOz: I thought  the version 2.14 and 2.1.5 only related to ROS1, whilst ROS2 just gets a "ros2" branch but upon investigation it seems the 2.1.5 tag applies to the ros2 branch too.  Will fix.

DingoOz:
all updated and ready for review thanks

firmware/battery-management-system/sdkconfig.debug
RandomSpaceship, line 1: Please remove whatever commit changed this file

flake.nix
RandomSpaceship: Since it's a perseus_sensors dependency now, you should remove it from here

software/ros_ws/patches.nix
```diff
@@ -18,6 +18,17 @@ let
...
       }
     );
+
+    rplidar-ros = rosPrev.rplidar-ros.overrideAttrs ({
+      version = "2.1.5"; # Version based on the ROS2 branch
+      src = final.fetchFromGitHub {
+        owner = "Slamtec";
+        repo = "rplidar_ros";
+        rev = "ros2"; # Using the ros2 branch
+        sha256 = "sha256-oNoDa+IqtQPe8bpfMjHFj2yx7jFUhfbIqaPRQCU/zMQ="; # need to find the right hash
```
RandomSpaceship, line 28: Please remove the unnecessary comments

software/ros_ws/src/perseus_sensors/package.xml
```diff
@@ -15,6 +15,7 @@
   <depend>sensor_msgs</depend>
   <depend>openssl</depend>
   <depend>nlohmann-json-dev</depend>
+  <depend>rplidar-ros</depend>
```
RandomSpaceship, line 18: This should be `exec_depend` not `depend` since you aren't using it in code, only at runtime manual launching or in launch files

RandomSpaceship (changes requested):

DingoOz:
I placed the necessary changes in the feat/perseus-lite branch, this PR is no longer needed.

PR #217, opened 2025-07-08
Fixed Livox Launch

UltraFishy:
- Dynamically updates config file with IP
- Needs to be tested
- Documentation has been added directly to launch file


software/ros_ws/src/perseus_sensors/launch/livox.launch.py
DingoOz, line 1: Could IP address detection get confused if we use the Wifi and the Ethernet on the Jetson Orin? Like if it works when wifi is turned off, is there an unknown chance the wrong host IP address is returned if the wifi is enabled? Maybe we could we add a param for specifying the interface so it is a bit more deterministic?

DingoOz (approved):
tested fine (although without access to livox hardware). Launch file looks correct and software standards compliant.

PR #219, opened 2025-07-11
Feat/perseus sensors/i2c imu driver

DingoOz:
This is an extension to the existing perseus_sensor package to allow use of imu's directly connected to the Orin (or Pi) via i2c.

Several models of imus are supported and addresses can be specified so multiple imus can be used. 

It is possible to daisy-chain STEMMA/QWIIC/PIICODEV imus on the i2c bus and use this package to read them all. Configuration is via the .yaml.

A new documentation page for using the i2c_imu node is included. 

Magic numbers have been avoided.

Tested on a Jetson Orin Nano.


DingoOz:
Please note that I have merged this into the feat/perseus_lite.

RandomSpaceship:
Why the close? It's quite valuable work, did you mean to convert it to a draft?

PR #248, opened 2025-08-20
docs(getting-started & documentation-index): Added distro and fixed docs

Lachlan-Ikeguchi:
Added a distrobution recomendation of Ubuntu 24.04.03 derivatives and
added back slashes to the end of the darkhttpd serve command's argument


docs/source/home/getting-started.md
DingoOz, line 1: Not sure what the basis for needing Ubuntu 24.04.03 LTS is. Is this based on a specific source or experience?   Generally the strength of using the nix managed environment is the support for almost all Linux distros and even WSL2.

Lachlan-Ikeguchi:
Screenshot 2025-08-20 at 16-21-16 Documentation - Perseus V2 documentation
Screenshot 2025-08-20 at 16-22-15 Getting Started - Perseus V2 documentation

docs/source/home/getting-started.md
Lachlan-Ikeguchi: I tried using Arch, Opensuse Tumbleweed and LMDE but I just couldn't seem to get the simulation to launch, but now moving to a Ubuntu 24.04.03 LTS derivative, it's finally working.  So I added a recommendation to suggest that the distro is the problem.

DingoOz: Are you able to to show what errors were being thrown by the non-Ubuntu distros?  We have implemented Nix to specifically avoid needing to force everyone onto the same Distro and thus far it has been really good at this.  If there is a common error for the non-Ubuntu distros it might just be a fix by adding a dependency to the flake.nix or similar.

RandomSpaceship:
This should not a distro-specific issue, and yes, the use of Nix is specifically to prevent needing to run a specific distro. It's likely something with Gazebo and/or graphics drivers, especially since it can and will freeze to silently download large assets in the background if a scene you loaded needs that. My guess would be that either newer GPU drivers got picked on Ubuntu 24, which is probably something that can be resolved by using different NixGL versions, _or_ it's an X11/Wayland thing, which can also be resolved but would need looking into.

docs/source/home/getting-started.md
DingoOz: I came across a bug whilst trying to fix this: https://github.com/ROAR-QUTRC/perseus-v2/issues/262

DingoOz:
I think we should split this into two parts:

One part to capture the changes that are not in question:
  - The nix build .#docs fix - this is a genuine improvement
  - The trailing slashes on the darkhttpd paths

Separate part (or just remove):
  - The Ubuntu 24.04 recommendation

As @RandomSpaceship mentioned, the simulation issues you experienced are likely related to NixGL/graphics drivers or X11/Wayland differences rather than the distro itself. Adding a distro recommendation would go against our goal of using Nix to support any Linux distribution.  I would recommend we go with the first part but leave the second part.

DingoOz (commented):
Looks good overall — the `nix build .#docs` fix is a genuine improvement and the trailing slashes are fine.

Minor point on the firewall note: "network must be set to trusted" is specific to `firewalld` zone terminology. A more general phrasing like "firewall rules must allow outgoing connections to download assets" would apply regardless of which firewall tool is in use. Not a blocker though.

DingoOz (approved):
LGTM

PR #252, opened 2025-08-28
Docs/autonomy 2026 team natalia

DingoOz:
Completed as per tutorial. Tested locally:
image

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+"cmake.sourceDirectory": "/home/natyh/perseus-v2"
+}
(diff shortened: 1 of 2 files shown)


TheFriendlyGoat (approved):

.vscode/settings.json
RandomSpaceship, line 1: We need to gitignore settings.json, I thought it already was but apparently not. Easiest way to fix is probably merge the PR then fix the gitignore and delete this file in the same commit.

RandomSpaceship (approved):
At some point the names need to be dot point formatted, when rendered on the site they all becomes one line and impossible to read

OMEN44 (approved):

PR #253, opened 2025-08-29
p-l-split-description: add Perseus Lite robot description package

DingoOz:
- Add URDF/xacro files for Perseus Lite robot model
  - Include 3D mesh files for visualization
  - Add ROS 2 control interfaces for simulation
  - Include launch files for viewing robot in RViz
  - Add Nix package definition"

- Robot URDF/xacro files with kinematic chain                                                                                               - High-quality 3D meshes for all robot components                                                                                             - RViz configuration files for visualization                                                                                                    - Launch files for robot state publisher                                                                                                        - Nix package definition for reproducible builds

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -318,7 +318,6 @@ qtcreator-*
 CATKIN_IGNORE
 
 .vscode/*
-!.vscode/settings.json
 !.vscode/tasks.json
 !.vscode/launch.json
 !.vscode/*.code-snippets
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,0 @@
-{
-  "cmake.sourceDirectory": "/home/natyh/perseus-v2"
-}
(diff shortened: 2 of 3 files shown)


RandomSpaceship (approved):
Looks good.

DingoOz:
> Overall, very good work. The biggest potential improvement I can see would be to try to re-use more files with the main `perseus_description` package, such as meshes. As long as it's marked as a dependency, you should be able to replace things like `file://$(find perseus_lite_description)/meshes/wheel.dae` with `file://$(find perseus_description)/meshes/wheel.dae` (the find command can work on any package that's present).

Yes I understand. I am also thinking that while this duplicates some files, keeping it separate makes the package easier for new team members to understand how P-L is captured... and P-L is likely where new team members will start. Also separation allows Perseus Lite to diverge independently over time without worrying about changes from the main Perseus package.

PR #255, opened 2025-08-29
fix: Exclude meshes from formatting

RandomSpaceship:
Exactly what it says on the tin. Should fix formatting issues for #253 and the like.


DingoOz:
Thanks for this, I suspect that beyond just Perseus-Lite this is will be good for Autonomy as we will need to improve some of the meshes that haven't been touched for a long time.

DingoOz (approved):

PR #256, opened 2025-08-29
Change naming standard from camel case to snake case

RandomSpaceship:
Currently our [software standards](https://roar-qutrc.github.io/standards/software.html) state to use `camelCase` naming (in C++, anyway). However, I've noticed that most ROS code uses `snake_case`, plus it's the Python language standard. I'm raising this because I don't like to mix and match it in a codebase if possible, and most of Nigel's code currently uses the `snake_case` style. Given the apparent de facto ROS standard, which I didn't know about at the time I wrote the standard, what's everyone's thoughts on changing it? It might be worth carving out an exception for firmware using Arduino, since that's `camelCase`, but then again, everything ESP-IDF is `snake_case`, so that would need to be a further discussion.

Vim regex replace to bulk-replace (re-run until no more changes are found):
`:%s/(?!^)\<([a-z_]+)([A-Z][a-z]+)/\1_\l\2/g`

Using `sed` to replace across the codebase (re-run several times for multi-word names, modified to hit all c/cpp/h/hpp files in `software/` and `firmware/`, mistakes fixed manually in diff):
`find software/ -name '*.cpp' -type f -exec sed -i -r -e 's/\<([a-z_]+)([A-Z][a-z]+)/\1_\l\2/g' -- {} +`

This PR should **not** be merged until:
1. There has been a discussion about whether it is useful
2. Several different people have approved it


OMEN44:
Thank you for clarifying and I do now see that you mentioned manually fixing errors in the original method. I'll have a look over it myself before approving later today hopefully.

docs/source/standards/software.md
```diff
@@ -586,13 +599,47 @@ Libraries may need internal, private header files - if necessary, these should b
...
+    int bar;
+};
+
+```
+
+#### Naming Convention (Arduino-only)
 
-Variables should be named in `camelCase`. The only exception to this rule is local static variables (that is, static within a function or class). These variables should be named in `snake_case` - this is to immediately signify to a programmer (who may not have seen the declaration) that they may behave differently to how they expect.
+The exception to the above if is you're writing firmware based on the Arduino core.
+If the code you're writing is primarily Arduino and/or FreeRTOS-based, feel, feel free to use either `camelCase` or `snake_case` - just keep whatever you do consistent internally in the project.
```
OMEN44, line 636: Duplicate word

RandomSpaceship:
Fixed the duplicate word. Also rebased onto most recent main.

firmware/battery-management-system/include/bq769x2.hpp
```diff
@@ -1359,27 +1359,27 @@ class bq76942
...
+        float min_cell_temperature;
+        float avg_cell_temperature;
         float cc3Current;
         float cc1Current;
         int32_t cc2Counts;
         int32_t cc3Counts;
     };
     struct da_status_6_t
     {
-        float accumulatedCharge;
```
OMEN44: Looks like numbers right before a capital have been missed

OMEN44 (approved):

GR3Y-SCALE (approved):

DingoOz (approved):
Great team work on this one.

PR #263, opened 2025-09-02
docs(simulation): Added the stamped=true param

DingoOz:
- the gazebo simulation requires the stamped param to be set to true
Tested locally and working: 
image


RandomSpaceship (approved):
Looks good, thanks for doing that.

PR #272, opened 2025-09-08
Fix/prod controller

UltraFishy:
I have added the generic_controller launch file to the nix apps, so that we don't use the old xbox controller node. I have updated the docs to reflect this change accordingly. 

\*insert doc image here\*

### DIFF
--- a/docs/source/home/perseus-operation/teleop.md
+++ b/docs/source/home/perseus-operation/teleop.md
@@ -167,7 +167,11 @@ nix run .#generic_controller -- type:=8bitdo wireless:=true
 Use a custom config file
 
 ```console
+# Absolute path
 nix run .#generic_controller -- config:=/absolute/path/to/my_controller.yaml
+
+# Alternative: Relative path (from current working directory)
+nix run .#generic_controller -- config:=./relative/path/to/my_controller.yaml
 ```
 
 ### Safe Operation


docs/source/home/perseus-operation/teleop.md
```diff
@@ -127,16 +127,47 @@ nix run .#ros2 -- run teleop_twist_keyboard teleop_twist_keyboard
...
+8BitDo controller, wireless
+
+```console
+nix run .#generic_controller -- type:=8bitdo wireless:=true
+```
+
+Use a custom config file
+
+```console
+nix run .#generic_controller -- config:=/absolute/path/to/my_controller.yaml
```
RandomSpaceship, line 170: A path relative to CWD should be fine too

software/ros_ws/src/perseus_input/launch/controller.launch.py
```diff
@@ -65,6 +65,7 @@ def generate_launch_description():
         name="generic_controller",
         output="both",
         emulate_tty=True,
+        # parameters=[{"config": config_path}],
```
RandomSpaceship, line 68:
```suggestion
```

software/ros_ws/src/perseus_input/src/generic_controller/generic_controller.cpp
```diff
@@ -10,6 +10,13 @@
 GenericController::GenericController(const rclcpp::NodeOptions& options)
     : Node("generic_controller", options)
 {
+    // // Output config path chosen
+    // this->declare_parameter<std::string>("config", "");
+
+    // // Log the config path
+    // auto config_path = this->get_parameter("config").as_string();
+    // RCLCPP_INFO(this->get_logger(), "Using config path: %s", config_path.c_str());
+
```
RandomSpaceship, line 19:
```suggestion
```
Because it's a ROS _node_ config file, the node doesn't actually know anything about the file, it only gets the parameters it provides.

RandomSpaceship (approved):
I'd forgotten this wasn't already done, thanks for the work.

RandomSpaceship (approved):

PR #285, opened 2025-11-06
Feat/perseus lite split/i2c imu driver

DingoOz:
This adds direct i2c device control to Perseus (or Perseus-Lite). 

Jetson Orins and Raspberry Pis can have direct GPIO connection to i2c devices, both supply 3.3V rails so with gnd SDA and SCL adding devices is pretty straightforward. 

Whilst Perseus probably doesn't want I2C IMUs directly connected in competition, the option is nice and if there are secondary Raspberry Pis it makes sense. 

This PR replaces a now closed (defunct) #219 .

========
Testing:

I have tested this with a LSM6DSOX IMU and the /imu/data topic is being published appropriately.

### DIFF
--- a/docs/source/home/perseus-operation/teleop.md
+++ b/docs/source/home/perseus-operation/teleop.md
@@ -140,9 +140,13 @@ The launch file accepts three parameters:
 
 | Argument   | Default   | Description                                                                                |
 | ---------- | --------- | ------------------------------------------------------------------------------------------ |
-| `type`     | `xbox`    | Controller type. Options: `xbox`, `8bitdo`.                                                |
+| `type`     | `xbox`    | Controller type. Options: `xbox`, `8bitdo`, `logitech`.                                    |
 | `wireless` | `true`    | Connection type. Set `true` for wireless, `false` for wired.                               |
 | `config`   | _(empty)_ | Path to a custom YAML config file. If provided, this overrides both `type` and `wireless`. |
+
+::: {note}
+The logitech controller config file has been configured for a logitech wired F310, and the 8bitdo controller config file has been configured for an 8BitDo Ultimate 2C controller. These same configuration files might not work for other models.
+:::
 
 Some example runs:
 
@@ -162,6 +166,12 @@ nix run .#generic_controller -- type:=xbox wireless:=false
 
 ```console
 nix run .#generic_controller -- type:=8bitdo wireless:=true
+```
+
+Logitech controller, wired
+
+```console
+nix run .#generic_controller -- type:=logitech wireless:=false
 ```
 
 Use a custom config file
--- a/software/ros_ws/src/perseus_input/launch/controller.launch.py
+++ b/software/ros_ws/src/perseus_input/launch/controller.launch.py
@@ -21,7 +21,7 @@ def generate_launch_description():
         DeclareLaunchArgument(
             "type",
             default_value="xbox",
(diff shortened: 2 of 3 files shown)


software/ros_ws/src/perseus_sensors/i2c_imu_driver/src/i2c_imu_hardware_interface.cpp
```diff
@@ -0,0 +1,400 @@
...
+            {
+                RCLCPP_WARN(rclcpp::get_logger("I2cImuHardwareInterface"),
+                            "Failed to configure optional I2C IMU sensor, continuing without IMU");
+                return hardware_interface::CallbackReturn::SUCCESS;
+            }
+        }
+
+        RCLCPP_INFO(rclcpp::get_logger("I2cImuHardwareInterface"),
+                    "Successfully configured I2C IMU device at %s:0x%02X",
+                    _i2c_bus_path.c_str(), _device_address);
```
RandomSpaceship: From my understanding of the ROS2 control design patterns, claiming hardware interfaces and setting up hardware should be in `on_activate`, not `on_configure`, so this code should be moved.

software/ros_ws/src/perseus_sensors/i2c_imu_driver/src/i2c_imu_hardware_interface.cpp
```diff
@@ -0,0 +1,400 @@
...
+                hardware_interface::StateInterface(
+                    info_.name, _state_interface_names[i], &_sensor_states[i]));
+        }
+
+        return state_interfaces;
+    }
+
+    void I2cImuHardwareInterface::_readLoop()
+    {
+        while (_reading_active.load())
```
RandomSpaceship: Why use a read thread instead of the main `read()` function?

software/ros_ws/src/perseus_sensors/i2c_imu_driver/src/i2c_device.cpp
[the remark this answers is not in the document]
DingoOz: Excellent thoughts. Updated so the Constructor does full initialisation and got rid of _is_initialized flag.

I kept isConnected() to support the 'required: false' use case for supporting optional IMUs. Kept it to attempt graceful degradation if an optional IMU gets unplugged rather than killing the node. Happy to remove if you'd prefer pure exception-based error handling throughout.

DingoOz:
>However, I noticed that the `i2c_imu_hardware_interface` implementation looks like it's an unused previous revision of what's in the standalone node. Was it meant to be included?

I shouldn't have included it. I have now removed the hardware_interface approach entirely in favour of a standalone I2cImuNode.

UltraFishy (approved):
I agree with what James brought up, it looks to be resolved though. Happy to merge!

SO-100_MOVEIT_INTEGRATION_PLAN.md
OMEN44: Is the `SO-100_MOVEIT_INTEGRATION_PLAN.md` meant to be included?

OMEN44 (changes requested):

SO-100_MOVEIT_INTEGRATION_PLAN.md
DingoOz: lol, no it was not meant to be on this branch :)  It is a surprise Christmas gift for you later :P

OMEN44 (approved):

PR #290, opened 2025-11-10
Increased TX Queue Length attribute of can network

BrockoliniMorgan:
Added a section to machine-setup.sh that creates a udev rules file to set the CAN network's txqueuelen attribute. I've tested it on the big-brain and it works as expected. Also added section to reload the udev rules so the change immediately takes effect.


RandomSpaceship (commented):
If you've tested and it works, all good. I would ask that you squash the git commits (merge them all into one commit) though, a change like this doesn't need 6 commits.

PR #305, opened 2025-11-21
Bugfix/rplidar package for using C1 and A3 rplidar with Perseus and P-L

DingoOz:
This fixes the "package not found" error that occurs when updating nix-ros-overlay to newer versions where rplidar-ros is not available for Jazzy.

rplidar-ros is not officially released for ROS 2 Jazzy in the rosdistro repository. Previously, patches.nix attempted to override rosPrev.rplidar-ros, but this package doesn't exist in the Jazzy distribution of nix-ros-overlay.


DingoOz:
Tested on actual hardware, C1 Lidar on Perseus Lite (Orin)

PR #306, opened 2025-11-22
Fix/controller timeout: Added timeout to controller node in case joy stops sending data

BrockoliniMorgan:
I added a 150ms timeout to the generic_controller node so if anything happens with the controller the rover won't keep moving off into the distance. The joy subscription is running at 10Hz, so it shouldn't get to the timeout unless something goes wrong. 

This error happened while testing today, and this change should fix it.


software/ros_ws/src/perseus_input/src/generic_controller/generic_controller.cpp
```diff
@@ -21,13 +21,43 @@ GenericController::GenericController(const rclcpp::NodeOptions& options)
...
     _actuatorPublisher = this->create_publisher<actuator_msgs::msg::Actuators>("bucket_actuators", 10);
-
+    _joyTimeoutTimer = this->create_wall_timer(JOY_TIMEOUT, std::bind(&GenericController::_joyTimeoutCallback, this));
+    _prevReceivedJoyTime = this->now();
     RCLCPP_INFO(this->get_logger(), "Generic controller initialized");
 }
 
+void GenericController::_joyTimeoutCallback(void)
+{
+    if ((_prevReceivedJoyTime.nanoseconds() - this->now().nanoseconds()) > 150000000)
```
DingoOz: Just checking... is this backwards the intent? Looks like it is subtracting received time from now() so that would always be a negative value?  Then .nanoseconds() returns a uint64_t which casts the number to unsigned which wraps the negative around to some large positive value?   So I assume despite it conceptually generating a negative value, in testing it is positive and large enough to pass the check?  So the fix would just be to subtract now() from received time?

DingoOz: Also we should not being using magic numbers like 150000000, best to define it as a variable and use that variable.

software/ros_ws/src/perseus_input/src/generic_controller/generic_controller.cpp
```diff
@@ -21,13 +21,43 @@ GenericController::GenericController(const rclcpp::NodeOptions& options)
...
+    {
```
DingoOz: If the timeout occurs, we should be logging something to the ROS logger. Otherwise it could really confuse a user if the trigger is firing for some unexpected reason but the log is silent so they don't know why the robot is not behaving as expected. Something like:

RCLCPP_WARN_THROTTLE(this->get_logger(), *this->get_clock(), 1000, "Joy timeout so publishing zero to twist commands to safely stop the rover");

Maybe the message could be better but you get the idea.

software/ros_ws/src/perseus_input/include/generic_controller/generic_controller.hpp
```diff
@@ -7,6 +7,8 @@
 #include <sensor_msgs/msg/joy.hpp>
 #include <string>
 
+#include "builtin_interfaces/msg/detail/time__struct.hpp"
```
DingoOz: I am pretty sure that this include is already provided by ROS so probably don't need to include it here.

software/ros_ws/src/perseus_input/include/generic_controller/generic_controller.hpp
```diff
@@ -7,6 +7,8 @@
 #include <sensor_msgs/msg/joy.hpp>
```
DingoOz: I have a concern that I think this will break Autonomous navigation. When Nav2 is running there are no received Joystick commands.   With twist_mux applied, Joystick gets a priority of 100 (high) and Navigation 10 (low). So if the joy stick code is writting zeros to twist messages these will clobber navigation messages causing Nav2 to not have any control.

DingoOz: I assume that it is likely because it is useful to be able to teleoperate the robot during Nav2 testing.  Nav2 testing usually starts with the rover not doing what we expect and needing to reset it. Being able to immediately teleoperate it is very useful.  Nav2 testing also usually involves re-playing a navigation over and over and it would be good to be able to rapidly teleoperate Perseus.

software/ros_ws/src/perseus_input/src/generic_controller/generic_controller.cpp
```diff
@@ -21,13 +21,43 @@ GenericController::GenericController(const rclcpp::NodeOptions& options)
         this->create_subscription<sensor_msgs::msg::Joy>("joy", 10, std::bind(&GenericController::_joyCallback, this, std::placeholders::_1));
     _twistPublisher = this->create_publisher<geometry_msgs::msg::TwistStamped>("joy_vel", 10);
     _actuatorPublisher = this->create_publisher<actuator_msgs::msg::Actuators>("bucket_actuators", 10);
-
+    _joyTimeoutTimer = this->create_wall_timer(JOY_TIMEOUT, std::bind(&GenericController::_joyTimeoutCallback, this));
+    _prevReceivedJoyTime = this->now();
     RCLCPP_INFO(this->get_logger(), "Generic controller initialized");
 }
 
+void GenericController::_joyTimeoutCallback(void)
```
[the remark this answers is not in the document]
BrockMorgan88: I added a timeout length parameter in the .yaml controller config file - we can change the timeout depending on what controller is connected because they might have different times between sending outputs

OMEN44 (approved):

PR #307, opened 2025-11-22
Feat/logitech controller

BrockoliniMorgan:
I added a config file for the logitech controller (currently our only controller until the new one arrives). I have tested on multiple devices, no drivers or firewall configuration needed on Ubuntu.

### DIFF
--- a/software/ros_ws/src/perseus_input/launch/controller.launch.py
+++ b/software/ros_ws/src/perseus_input/launch/controller.launch.py
@@ -44,7 +44,10 @@ def generate_launch_description():
     is_xbox = EqualsSubstitution(controller_type, "xbox")
     is_logitech = EqualsSubstitution(controller_type, "logitech")
     controller_configs = {
-        "xbox": ["xbox_controller", IfElseSubstitution(is_wireless, "_wireless.yaml", "_wired.yaml"),],
+        "xbox": [
+            "xbox_controller",
+            IfElseSubstitution(is_wireless, "_wireless.yaml", "_wired.yaml"),
+        ],
         "logitech": ["logitech_controller.yaml"],
         "8bitdo": ["8bitdo_controller.yaml"],
     }


docs/source/home/perseus-operation/teleop.md
```diff
@@ -141,6 +141,7 @@ The launch file accepts three parameters:
 | Argument   | Default   | Description                                                                                |
 | ---------- | --------- | ------------------------------------------------------------------------------------------ |
 | `type`     | `xbox`    | Controller type. Options: `xbox`, `8bitdo`.                                                |
+| `type`     | `xbox`    | Controller type. Options: `xbox`, `8bitdo`, `logitech`.                                    |
```
DingoOz: Thanks for updating the docs. Just wondering if there is a version constrain for this logitech? I am not sure if there are different versions that logitech sell. If it is a specific gamepad, it would be even better to specify which one is being used and supported.

software/ros_ws/src/perseus_input/launch/controller.launch.py
```diff
@@ -42,13 +42,19 @@ def generate_launch_description():
...
+    is_logitech = EqualsSubstitution(controller_type, "logitech")
     xbox_controller_config = [
         "xbox_controller",
         IfElseSubstitution(is_wireless, "_wireless.yaml", "_wired.yaml"),
     ]
     eightbitdo_controller_config = ["8bitdo_controller.yaml"]
+    logitech_controller_config = ["logitech_controller.yaml"]
     controller_config_name = IfElseSubstitution(
-        is_xbox, xbox_controller_config, eightbitdo_controller_config
+        is_xbox,
```
DingoOz: This is ok for now but if we need to add any more then I would suggest we turn this into a list, something like: controller_configs = {
      'xbox': xbox_controller_config,
      'logitech': logitech_controller_config,
      '8bitdo': eightbitdo_controller_config,
  }

DingoOz (approved):

PR #316, opened 2025-11-25
Updates the P-L Quickstart from xbox_controller to generic_controller

DingoOz:
Recent merge broke the documentation that referred to xbox_controller and not now generic_controller. 

Tested and working: 
image


UltraFishy (approved):
Tough stuff 💪

PR #321, opened 2025-11-27
Feat/web UI/bucket widget

OMEN44:
This PR includes a bucket drive widget similar to the web drive widget. This widget has all the same timeout protection methods that the teleop widget has and has very similar internal operations.

This widget uses the default settings for the bucket used in the controller configs and it's outputs have been verified using `ros2 topic echo` and compared with the outputs of an existing controller.

User interface
image
Settings defaults
image

Additionally, this PR includes mobile styling for the web-ui that collapses the top menu bar into two dropdown menus preventing horizontal scrolling.

Finally, this PR includes updates to the logitech config that were found to be incorrect during testing.


software/web_ui/src/lib/widgets/imuViz.svelte
```diff
@@ -0,0 +1,107 @@
+<script lang="ts" module>
+	// This is to expose the widget settings to the panel. Code in here will only run once when the widget is first loaded.
+	import type { WidgetGroupType, WidgetSettingsType } from '$lib/scripts/state.svelte';
+
+	export const name = 'IMU Visualizer - 3D';
+	// These properties are optional
+	export const description = 'Visulaise IMU data with a 3D rover model';
```
DingoOz: Typo on Visualise

software/web_ui/src/lib/widgets/imuViz2D.svelte
```diff
@@ -0,0 +1,132 @@
+<script lang="ts" module>
+	// This is to expose the widget settings to the panel. Code in here will only run once when the widget is first loaded.
+	import type { WidgetGroupType, WidgetSettingsType } from '$lib/scripts/state.svelte';
+
+	export const name = 'IMU Visualiser - 2D';
+	// These properties are optional
+	export const description = 'Visulaise IMU data with a 3D rover model';
```
DingoOz: Is "3D" a typo, should be 2D?

DingoOz: Same typo with Visualise

software/web_ui/src/lib/widgets/imuViz.svelte
DingoOz: imuViz and imuViz2D are lacking the ROS initialisation. Should they be part of this PR, or isolated to a separate branch/PR so they can be worked on separately to make work?

...noting recently merged PRs support direct GPIO IMUs now, which might make testing easier than needing the m2m2 or livox running.

software/web_ui/src/lib/components/connection-menu-items.svelte
```diff
@@ -0,0 +1,68 @@
...
+	import Input from './ui/input/input.svelte';
+	import { OpenInNewWindow } from 'svelte-radix';
+	import { getRosConnection } from '$lib/scripts/ros-bridge.svelte';
+
+	let {
+		toggleConnection,
+		changeAddress,
+		address,
+		disableServerIP,
+		openAddressMenu,
```
DingoOz: Looks like openAddressMenu is received as a prop but is modified locally. Can't see the needed bind: or callback to pus any local changes to the parent.

software/web_ui/src/lib/components/connection-menu-items.svelte
```diff
@@ -0,0 +1,63 @@
...
+	import Button from './ui/button/button.svelte';
+	import Label from './ui/label/label.svelte';
+	import Input from './ui/input/input.svelte';
+	import { OpenInNewWindow } from 'svelte-radix';
+	import { getRosConnection } from '$lib/scripts/ros-bridge.svelte';
+
+	let { toggleConnection, changeAddress, address, disableServerIP, customAddress, domainId } =
+		$props();
+
+	let openAddressMenu = $state(false);
```
DingoOz: Still looks like openAddressMenu is declared locally and is not sync'd wit the parent. The parent component appears to have its own openAddressMenu state. So two separate states.

software/web_ui/src/lib/widgets/bucketDrive.svelte
```diff
@@ -0,0 +1,254 @@
...
+
+	const deadBand = (value: number, threshold: number) => {
+		return Math.abs(value) < threshold ? 0 : value;
+	};
+
+	const publishActuatorValues = () => {
+		if (!topic) return;
+
+		const velocities = [
+			deadBand(handles.lift.value, 0.05) *
```
DingoOz: Deadband value is hardcoded multiple times here.

DingoOz (approved):

software/web_ui/src/routes/+page.svelte
```diff
@@ -128,24 +129,51 @@
...
 	};
+
+	let isMobile = $state<boolean>(false);
+
+	onMount(() => {
+		// check if the device is mobile
+		if (/Mobi|Android|iPhone|iPad|iPod|BlackBerry|Windows Phone/i.test(navigator.userAgent)) {
+			isMobile = true;
+		}
+		console.log('isMobile:', navigator);
```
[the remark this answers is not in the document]
OMEN44: This one is intentional as mobile detection is browser/device dependent so it's much easier to check the output when something is wrong than to add it and try and reproduce. That onMount should also be evaluated only once per page refresh so there is no performance loss.

software/web_ui/src/lib/components/connection-menu-items.svelte
OMEN44: The state of each separate menu never needs to be sync'd since connection-menu-item.svelte is the mobile view and the ui shouldn't be able to switch between desktop and mobile. The reason that connection-menu-item.svelte uses functions and variables from it's parent is because it has the same buttons. That is except for openAddressMenu which is used to programmatically open a dialog from a custom trigger.

I will rename this file as connection-menu-item.svelte is not very descriptive of what it actually is.

software/web_ui/src/lib/widgets/bucketDrive.svelte
OMEN44: Good catch. I'll add a constant.

OMEN44 (commented):
- Renamed connection-menu-items.svelte to connection-menu-mobile.svelte.
- Added missing timeout handle
- Added missing cleanup operations

DingoOz (approved):
Great work @OMEN44

PR #324, opened 2025-11-28
Feat/bucket launch: Added a flag to perseus launch file to run the payloads

BrockoliniMorgan:
I added a 'payload' flag to the perseus launch file so we can run a payload's launch file at the same time - means that we don't have to open up multiple extra panes, go into nix shells, and run the nodes individually, saving time. I made the bucket launch file as well - passes the required parameters to the nodes, which were already being passed to the perseus launch file anyways. 
It will be easy enough to add another launch file when we get the post landing arm or space resources up and running too.

### DIFF
--- a/software/ros_ws/src/perseus/launch/perseus.launch.py
+++ b/software/ros_ws/src/perseus/launch/perseus.launch.py
@@ -23,7 +23,6 @@ def generate_launch_description():
     use_mock_hardware = LaunchConfiguration("use_mock_hardware")
     hardware_plugin = LaunchConfiguration("hardware_plugin")
     can_bus = LaunchConfiguration("can_bus")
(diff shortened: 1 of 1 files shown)


software/ros_ws/src/perseus/launch/perseus.launch.py
```diff
@@ -23,6 +23,7 @@ def generate_launch_description():
     use_mock_hardware = LaunchConfiguration("use_mock_hardware")
     hardware_plugin = LaunchConfiguration("hardware_plugin")
     can_bus = LaunchConfiguration("can_bus")
+    # payload = LaunchConfiguration("payload")
```
OMEN44: Should this commented line be included?

OMEN44 (approved):

PR #342, opened 2025-12-06
Feat(templates): Added basic Ros2 template and docs

ZephCreations:
## Summary

Adds new template code for creating Ros2 packages and nodes. Includes documentation and CMake snippets for easier and streamlined use.

## What Changed

- Added Ros2 template package
- Added Docs describing templates and their use
- Added CMake snippets for custom Cmake code

## Why

Added to provide assistance to newcomers for creating their first package and experimenting with Ros, providing a firmer understanding. Also ensures packages are standardised and maintain the same format. Provides a reference as well for running, creating, and maintaining a package. An additional feature would be moving the CMake snippet description from the README.md file to docs.

## Testing

- [x] Unit tests pass
- [x] Integration tests pass
- [x] Manual testing completed
- [x] Documentation updated

### Test Instructions

1. Follow the instructions in the docs and copy the template code into the ros_ws folder
2. It should all build and run fine with no issues
3. Mainly doc updates and template code written, so no major testing required apart from basic functionality (as that is all it provides)

## Breaking Changes

- None

### DIFF
--- a/software/ros_ws/nix-packages/perseus-simulation.nix
+++ b/software/ros_ws/nix-packages/perseus-simulation.nix
@@ -10,6 +10,9 @@
   ros-gz-image,
   ros-gz-interfaces,
   ros-gz-sim,
+  rosbridge-server,
+  twist-mux,
+  twist-stamper,
   yaml-cpp-vendor,
 }:
 buildRosPackage rec {
@@ -31,6 +34,9 @@ buildRosPackage rec {
     ros-gz-image
     ros-gz-interfaces
     ros-gz-sim
+    rosbridge-server
+    twist-mux
+    twist-stamper
   ];
   nativeBuildInputs = [ ament-cmake ];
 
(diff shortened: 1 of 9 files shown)


docs/source/development/software/templates.md
DingoOz: This info does not appear to address the nix environment in which development takes place.  At minimum the user should be directed to enter the nix develop shell first and then as per https://roar-qutrc.github.io/systems/software/core.html the user should be told to least stage the code before running the /scripts/nix-package.sh, then running colcon build.

Also with this, the . install/setup.sh should be unnecessary.

docs/source/development/software/templates.md
```diff
@@ -0,0 +1,72 @@
+# Templates
+
```
DingoOz: This document does not like to other background documents that explain what a package is and what you would use it for. Nor does it seem to try to inform the reader of this. 

Suggest in the intro some treatment of what a package is and what you do with it, or a direct like to the ROS2 Jazzy tutorials or elsewhere that addresses this

PR #346, opened 2025-12-23
Feat/curl setup script

BrockoliniMorgan:
I added a setup script that can be run using one line to automate the getting started process. I tested it on a live usb of ubuntu, but haven't tested on any other distros. The script should work the same on all debian-based distros. I've changed the docs so this is the recommended method, but the old method is still in there for other distros (or someone who wants to do it manually).

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -108,6 +108,31 @@
             pkg:
             builtins.elem (pkgs.lib.getName pkg) [
               "drawio"
+              # CUDA packages
+              "cuda_cudart"
+              "cuda_nvcc"
+              "cuda_cccl"
+              "libcublas"
+              "libcufft"
+              "libcurand"
+              "libcusolver"
+              "libcusparse"
+              "libnpp"
+              "cuda_nvrtc"
+              "cuda_nvml_dev"
+              "cuda_profiler_api"
+              "cudatoolkit"
+              "libnvjitlink"
+              "cuda_cuobjdump"
+              "cuda_gdb"
+              "cuda_cuxxfilt"
+              "cuda_nvdisasm"
+              "cuda_nvprune"
+              "cuda_sanitizer_api"
+              "cuda_nvtx"
+              # ONNX Runtime CUDA dependencies
+              "cudnn"
+              "cudnn-frontend"
             ];
         };
         # we don't need to apply overlays here since pkgs-unstable is only for pure python stuff
@@ -130,6 +155,7 @@
             nodejs_23
             yarn
             nixgl-script
+            nixcuda-script
             ncurses
             glibcLocales
             yaml-cpp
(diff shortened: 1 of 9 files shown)


software/scripts/member-setup.sh
```diff
@@ -0,0 +1,57 @@
...
+fi
+
+# Run the nix-setup script
+echo "Running nix-setup.sh script. If asked, accept all config options by typing 'y', then press enter."
+
+cd ~/perseus-v2
+./software/scripts/nix-setup.sh
+
+# Check the script ran correctly and didn't fail
+EXIT_STATUS=$?
```
DingoOz: Is $? always 0 here because line 5 sets -euo pipefail so it already exits on failure?

Should we just remove the check  or disable errexit via "set +e" before the command to allow the script give the user a custom message?

software/scripts/member-setup.sh
```diff
@@ -0,0 +1,57 @@
...
+
+if ! [ $EXIT_STATUS -eq 0 ]; then
+  echo "nix-setup.sh script failed (Exit code $EXIT_STATUS). Exiting"
+  exit
```
DingoOz: As above, should this be "exit 1" not "exit"?

software/scripts/member-setup.sh
```diff
@@ -0,0 +1,57 @@
...
+fi
+
+# Build nix packages
+echo "Building nix packages. If asked, accept all config options with 'y'"
+/nix/var/nix/profiles/default/bin/nix build --accept-flake-config
+
+echo "Setup script ran successfully!"
+echo "Restarting shell"
+
+exec $SHELL
```
DingoOz: should 

exec $SHELL

be:

exec "$SHELL" 

to handle paths with spaces? I assume this would break on paths with spaces otherwise.

DingoOz (changes requested):

software/scripts/member-setup.sh
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env bash
+
+# If you're changing this file, make sure to read the systems/software/scripts docs
+
+set -euo pipefail
+
+# exit if run as root
+if [ "$EUID" -eq 0 ]; then
+  echo "Please run as yourself! Running as superuser (ie, with sudo) breaks the setup."
+  exit
```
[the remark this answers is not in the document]
BrockMorgan88: Just changed the exit code. I also changed the exit code of the machine setup script to one when it fails,

software/scripts/member-setup.sh
BrockMorgan88: Good point. Removed the error checking - the nix-setup script should give proper error messages

PR #352, opened 2026-01-03
Feat/map gui:  improving waypoint extraction 

StarEater271:
created a basic mapping feature which creates waypoints and creates yaml file, using an image that has been saved into ros_ws/map_gui/map_gui


DingoOz:
Looking really good. 
image
Is it possible to add a number to each waypoint (or label each on the map with the text label) so it is immediately clear the order in which they are captured? Also on my screen the yellow and red  indicators are very small, can the size be larger (or adjustable)?  Also the Hex colour column did not seem to have any impact on the preview screen. 

This is going to be a great tool for comp! :)

DingoOz:
Also, we need some way to set the origin of the map (another mode, or a new type of waypoint). This origin will impact the coords for all other waypoints too.   There also needs to be a scale function. Current the yaml is in pixels but it needs to be in meters, so either a a text input for pixel to meters or a drawing tool or similar is needed.

PR #368, opened 2026-01-13
Feat: Add initial message and service definitions for perseus_msgs package

bocho0600:
Add a perseus_msgs package to save msg and srv for:
- Autonomy WebUI interfaces
- Object Detection msg


bocho0600:
> Since this package was made I've realised it makes more sense to call is `perseus_interfaces`. This follows the official ros2 tutorial and msgs might be misleading as this package also contains services.

@OMEN44 fixed the name, can you review it again

PR #377, opened 2026-01-17
Adding CUDA support on PC and Orin to our nix environment

DingoOz:
## Summary
add Nix-managed CUDA toolkit support for portable GPU development
add `nixcuda` wrapper for running CUDA binaries with GPU passthrough
add ONNX Runtime with CUDA support for ML workloads
add cuda-test.sh and cuda-demo.sh for verification and demonstration

This does have sandbox = "relaxed" in the flake.nix  for documentation purposes (this is network access during build). 

Tested on both PC and Orin.

Not yet cached via cachix so is built from source. (needs about 25Gb swap on normal PCs and will take an hour or so to build). 

Have tested on PC and Orin. All GPUs detected and shown working. 

There is the /software/scripts/cuda-test.sh and /software/scripts/cuda-demo.sh to help with testing.

https://github.com/user-attachments/assets/a4ca0de8-c9e0-4386-8495-8140f0fe9aca

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -384,13 +384,8 @@
     };
   nixConfig = {
     extra-substituters = [
-      "https://perseus-v2.cachix.org"
       "https://roar-qutrc.cachix.org"
       "https://ros.cachix.org"
-    ];
-
-    extra-trusted-public-keys = [
-      "perseus-v2.cachix.org-1:Ey7Kc4bRmGHCfsT20RvTH9o2DNUImxjsENlyI96z4zI="
     ];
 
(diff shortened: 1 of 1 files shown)


DingoOz:
Have uploaded the x86 and Arm8 built derivatives to the new parallel cache now. Builds and tests should be significantly faster

DingoOz:
I have removed the parallel cachix cache reference. This still works because code once existed here that caused built derivatives to be pulled in and the CI/CD workflows copied them to the main cache.

RandomSpaceship (approved):
Looks good. If you've tested it, I'm happy with it.

PR #354, opened 2026-01-08
Feat/nix update

RandomSpaceship:
Update `nix-ros-workspace` input, disable docs sandboxing to reduce maintenance burden.

PR #378, opened 2026-01-17
Fix/trusted users

BrockoliniMorgan:
The nix update recently (PR #354) set the sandbox to 'relaxed'. This setting is ignored on most devices due to the user not being in the trusted-users for nix. I updated the nix-setup.sh script to add this. I'll also put a message out to members to re-run the script upon seeing this warning.
screenshot


DingoOz (approved):
LGTM. The approach looks sound.

PR #379, opened 2026-01-18
fix: Docs sandbox issue

BrockoliniMorgan:
For the workflow:
Added trusted-uris for each of the intersphinx-mapping configs, set the option for relaxed sandbox
For regular code:
Changed the sandbox back to full, and removed __noChroot for the docs derivation


docs/default.nix
```diff
@@ -95,10 +95,6 @@ let
         ;
     };
 
-    # disable sandboxing so sphinx extensions can grab their internet resources
-    # (intersphinx, tippy, etc)
-    __noChroot = true;
-
```
RandomSpaceship: Setting `sandbox` to relaxed makes no difference if you don't leave this here. Anybody building the docs locally should already be using `nix develop` and `uv run make html` since it's so much faster.

BrockMorgan88: So `nix build .#docs` should only be used during the Github Actions? The current [Documentation Development](https://roar-qutrc.github.io/development/documentation-index.html) page doesn't reflect that. I can update it to recommend only using `nix develop .#docs` instead.

PR #387, opened 2026-01-26
feat: Add occupancy viewer widget for displaying SLAM map with robot pose and path

bocho0600:
- Adding web widget for slam map , nav2 planned path and robot position.
image


software/web_ui/src/lib/widgets/occupancyView.svelte
```diff
@@ -0,0 +1,311 @@
...
+					description: 'Marker size (base scale in pixels)',
+					value: 2
+				}
+			},
+			path: {
+				topic: { type: 'string', value: '/plan' },
+				messageType: { type: 'string', value: 'nav_msgs/msg/Path' }
+			}
+		}
+	});
```
OMEN44: Do u have any typescript extensions in vs code? There are type errors from using "string" instead of "text"

OMEN44 (changes requested):
There are several other type errors that you should fix. I would recommend this VS code plugin: 
`svelte.svelte-vscode`

I will add this to the list of recommended plugins for the project in a future PR.

PR #389, opened 2026-01-29
chore: Update `nix-direnv` version

RandomSpaceship:


.envrc
```diff
@@ -1,7 +1,7 @@
 #! /usr/bin/env bash
 # use nix direnv to make evaluation faster
-if ! has nix_direnv_version || ! nix_direnv_version 3.0.5; then
-  source_url "https://raw.githubusercontent.com/nix-community/nix-direnv/3.0.5/direnvrc" "sha256-RuwIS+QKFj/T9M2TFXScjBsLR6V3A17YVoEW/Q6AZ1w=" &>/dev/null
+if ! has nix_direnv_version || ! nix_direnv_version 3.0.7; then
+  source_url "https://raw.githubusercontent.com/nix-community/nix-direnv/3.0.7/direnvrc" "sha256-RYcUJaRMf8oF5LznDrlCXbkOQrywm0HDv1VjYGaJGdM=" &>/dev/null
```
DingoOz: I think the 3.0.7 readme has a the wrong hash in it.  Looks like there was a commit to 3.0.7 tag after the Readme was commited that updated the fetchurl checksum: [](https://github.com/nix-community/nix-direnv/commit/70e5334  )

The correct hash is "sha256-bn8WANE5a91RusFmRI7kS751ApelG02nMcwRekC/qzc="

Given most of us already have nix-direnv installed, the source_url line doesn't execute so testing will succeed. It goes straight to using the flake so the hash doesn't get checked but fresh installs will fail the hash check:

```
dingo@ml-server:~/perseus-v2$ direnv fetchurl "https://raw.githubusercontent.com/nix-community/nix-direnv/3.0.7/direnvrc" "sha256-RYcUJaRMf8oF5LznDrlCXbkOQrywm0HDv1VjYGaJGdM="  
direnv: error hash mismatch. Expected 'sha256-RYcUJaRMf8oF5LznDrlCXbkOQrywm0HDv1VjYGaJGdM=' but got 'sha256-bn8WANE5a91RusFmRI7kS751ApelG02nMcwRekC/qzc='
```

DingoOz (changes requested):

PR #391, opened 2026-02-03
Flake Update

BrockoliniMorgan:
Summary of changes:
- Updated flake inputs
- Packaged freeimage from source, because it has been abandoned and is no longer in nixpkgs
- Linked libjpeg to freeimage statically so the patches can take effect and not mess with other packages
 - Gz-sim-vendor tries to link freeimage, but it links to a different version of libjpeg in the environment, causing the libjpeg patches to not matter
 - Libjpeg is a dependency for *many* packages, so we can't just override it at the base without building hundreds of packages from source
- Updated home manager to 25.11, and locked the flake input to be the same version as our nixpkgs input
 - Updated all the config files to be the updated options for 25.11
- Updated all CMake minimum version requirements to 3.10 so we don't get a warning that below 3.10 is deprecated
- Updated colcon packages with pyproject and setuptool
- Removed outdated patches like librealsense, rplidar, jointstatepublisher, rosbridge-library, gst-plugins-rs
- Added patches for gz-vendors because the ament-cmake-vendor-package-wrapped in nix-ros-overlay has changed versions for some packages (the vendored packages that are exposed to the vendors), but the vendors still search for the previous version, so they don't find the package
- Added XTensor and XTL versions for nav2-mppi-controller. The ones in nixpkgs are the unstable versions, but nav2-mppi-controller needs the stable versions
- Updated the PATCHES.md file with a proper description of each patch I've made
- Updated our hardware plugins. One of the previous functions used was deprecated, so the variable type needed to change to a struct that contains the variable we used previously
- Updated the formatter and typo checker - All the changes to our .md files are because of this.
 - Added an ignore word for 'ROS' to the typo checker - the new version is case insensitive so ros -> ROS was picking up ROS and trying to correct it to ROS

### DIFF
--- a/software/home-manager/cli-programs/git.nix
+++ b/software/home-manager/cli-programs/git.nix
@@ -2,19 +2,7 @@
 {
   programs.git = {
     enable = true;
-    delta = {
-      enable = true;
-      options = {
-        # https://dandavison.github.io/delta/hyperlinks.html
-        hyperlinks = true;
-        hyperlinks-file-link-format = "vscode://file/{path}:{line}";
-        navigate = true;
-        side-by-side = true;
-        line-numbers = true;
-        true-color = "always";
-      };
-    };
-    extraConfig = {
+    settings = {
       diff = {
         color-moved = "default";
       };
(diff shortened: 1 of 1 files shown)


UltraFishy (approved):
Good work Brock!

RandomSpaceship:
Aside from the `typos.toml` (adding the ROS ignore regex completely disables the correction) it looks good. Happy for it to be merged and fix that later.

BrockMorgan88:
> Aside from the `typos.toml` (adding the ROS ignore regex completely disables the correction) it looks good. Happy for it to be merged and fix that later.

Looks like it works when testing
image

PR #393, opened 2026-02-09
Added Unifi config

dbreth:
- Added the unifi config to ./software/network/
- It contains the static IP's of:
  - PoE Switch
  - U7 Long range
  - U7 Antenna
  - Livox Lidar

- It has the proper configuration for the U7 Long Range to only connect to the U7 Antenna.
- The DHCP range has changed to .50 to .254 to make sure there is no IP collisions with set static IPs


DingoOz (changes requested):
Great work, this is really needed. To help avoid the CI erros, can .unifi file ideally have a .gitattributes entry marking it as binary (*.unifi binary) to prevent git from attempting text diffs/merge on it.

Also should we list the static IP addressed devices and their IPv4 addresses in a small readme or other form of documentation?

PR #394, opened 2026-02-10
Feat/web UI/save videos

BrockoliniMorgan:
I added a setting to the camera widget to allow for saving video. It works by splitting the gstreamer pipeline in two - one part goes to the webui, the other goes to the file. It also has some queues so the splitter doesn't slow down the other side, with writing to the file taking too long, so the webui can't get another frame etc.
You can test it out on your laptop webcam. It saves an avi file, which can be played using vlc. Currently, it only saves to the device which is running the camera,which can then be copied onto a usb when needed. I've tested on the Jetson and on my laptop.


UltraFishy (approved):
Apart from the jpeg encoding, I like it

PR #396, opened 2026-02-21
Chore/web UI/updates

OMEN44:
This branch updates all the broken/outdated dependencies used by the Web UI.
Changes:
- Node has been removed since yarn ships with it's own node install
- All the shadcn-svelte components used have been updated. These are in the `software/web_ui/src/lib/components/ui` folder and can be skipped as I have not changed anything inside them.
- ROSLIB
  - New version now has very strict types. There are new type definitions inside the lib/scripts folder and all widgets have been updated so they still work with the library
- TopicMonitor and NodeMonitor have been updated with the new JsonTree.svelte component


software/web_ui/src/lib/components/settingsForm.svelte
```diff
@@ -9,11 +9,13 @@
...
 	let { layout } = $props();
 
+
+
 	const toggleLock = () => {
 		repo(Layout).update(layout.id, { isLocked: !layout.isLocked });
 	};
 
-	let settingsForm = $state<{ title: string; description: string }>({
+	let settingsForm = $derived<{ title: string; description: string }>({
```
DingoOz: settingsForm being changed from $state to $derived but used with bind:value needs a writable value, so this change might break editing layout titles and descriptions. Should it change back to $state (and sync from props via $effect if needed)?

software/web_ui/src/lib/widgets/nodeViewer.svelte
```diff
@@ -93,38 +160,15 @@
 	</div>
 
 	<div class="mt-2 h-[calc(100%-40px-0.5rem)] w-full">
-		{#if nodeData}
+		{#if communicationsTree}
```
DingoOz: Is {#if communicationsTree} always true because the derived value returns {} even when no node is selected? Could we change the condition to check selectedNode or nodeData instead?

software/web_ui/src/lib/widgets/busPowerManager.svelte
```diff
@@ -128,7 +126,7 @@
 		<p>Waiting for power bus data</p>
 	{/each}
 </div>
-<p class="ml-2 opacity-35">FAULT on the drive bus likely indicates the drive stop is engaged.</p>
+<p class="ml-2 opacity-35">Note: FAULT on the drive bus likely indicates the drive stop is engaged.</p>
 
 <style>
 	:global(.power-button) {
```
DingoOz: Is this CSS rule now dead code as the class was removed.

software/web_ui/src/lib/widgets/webrtc/videoWrapper.svelte
```diff
@@ -33,11 +33,11 @@
 		isSettingsOpen = !isSettingsOpen;
 	};
 
-	let resolution = $state<string>(
+	let resolution = $derived<string>(
```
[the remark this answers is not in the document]
OMEN44: Same as above, this makes the variables update when props change

software/web_ui/src/lib/components/settingsForm.svelte
OMEN44: Svelte allows for derived variables to bound to props and they can be *temporarily* written to. This change from $state to $derived is to address a compiler error about the variable not updating when it's dependencies change as a $state rune.

software/web_ui/src/lib/widgets/busPowerManager.svelte
OMEN44: I'm not sure what CSS rule your referring to?

PR #400, opened 2026-02-28
Add IMU Bias Estimator and Remover components with launch files

bocho0600:
- Implemented BiasEstimator and BiasRemover as composable nodes for IMU bias correction.
- Created launch files: bias_remover.launch.py, lidar_processing.launch.py, pcl_to_lsc.launch.py, rs_launch.py, topic_remapper_launch.py.
- Added topic remapper node to dynamically remap topics with configurable parameters.
- Integrated RealSense camera launch with image compression capabilities.
- Enhanced bias estimator with stationary detection logic and rate control.
- Implemented topic remapper to publish messages at a reduced frequency.

The custom nodes can be either tested in simulation or real-life with 3D lidar:

```bash
ros2 launch perseus_sensors lidar_processing,launch.py use_sim_time:=false #if tested in real-life
```

### DIFF
--- a/software/ros_ws/src/perseus_sensors/launch/lidar_processing.launch.py
+++ b/software/ros_ws/src/perseus_sensors/launch/lidar_processing.launch.py
@@ -20,8 +20,7 @@ Launch Arguments:
 """
 
 from launch import LaunchDescription
-from launch.actions import DeclareLaunchArgument, IncludeLaunchDescription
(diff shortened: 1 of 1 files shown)


UltraFishy (commented):
I would like to see a testing method of all of these launch files before merging to main.
What needs to be running and when. What sensors need to be connected, etc.

PR #404, opened 2026-03-03
feat: Added and patched FAST-LIO ROS2 package

BrockoliniMorgan:
FAST-LIO for ROS2 sets the cpp std version to 14 (in multiple places). Jazzy expects std version 17, so I patched it for @LucaLubrano :)
Test using:
`nix develop .#pkgs.ros.fast-lio`


BrockMorgan88:
@DingoOz You might need to add it to the devShell in the flake. We had some issues testing using `nix develop .#pkgs.ros.fast-lio`

PR #408, opened 2026-03-10
Feat/fast lio mapping

LucaLubrano:
Added patches to fast-lio to work on our rover in jazzy

Fixed algorithms so that fast-lio works in our simulation environment

Integrated and tested implementation in both simulation and on perseus


software/ros_ws/src/perseus_mapping/config/livox_mid360.yaml
```diff
@@ -0,0 +1,48 @@
+/**:
+  ros__parameters:
+    feature_extract_enable: true
+    point_filter_num: 3
+    max_iteration: 3
+    filter_size_surf: 0.5
+    filter_size_map: 0.5
+    cube_side_length: 1000.0
+    runtime_pos_log_enable: false
+    map_file_path: "/home/luca-lubrano/maps/scan.pcd" # luca personal file path
```
UltraFishy: Another file path please

PR #409, opened 2026-03-10
Rover Control Board

BrockoliniMorgan:
I've refactored the [Artemis RCB code](https://github.com/qut-robotics-club/ARC_Artemis/tree/main/Firmware/rover_control_board) to work with our hi-can library. It's been a while since I wrote it, so I'm going back through it over the next few days, with testing starting tomorrow. This is still a draft, because as of now it's untested, but any feedback is appreciated. I'll probably also need to go through the config for precharge values over the next few days to work with our new voltage regs - [YH11060D](https://rhoelect.gr/product/yh11060d-step-down-dc-dc-converter/) - so any advice for that is also welcome :)


DingoOz (changes requested):
I have flagged some potentially critical items.

firmware/rover-control-board/src/main.cpp
```diff
@@ -0,0 +1,304 @@
...
+            packet_manager->set_transmission_config(
+                flagged_address_t(standard_address_t(RCB_DEVICE_ADDRESS, static_cast<uint8_t>(id), static_cast<uint8_t>(hi_can::addressing::power::distribution::rover_control_board::power_bus::parameter::POWER_STATUS))),
+                power_bus.get_transmission_config());
+        }
+        catch (const std::exception& error)
+        {
+            printf("Error \"%s\" while setting parameter groups for %s", error.what(), name.c_str());
+            return;
+        }
+    }
```
RandomSpaceship: You shouldn't just return from `app_main` if an error occurs, since that leaves the user with no indication of what's happened. A better option would be to go into an infinite loop error handler function which rapidly blinks the LED to show that it's in an error state, potentially before rebooting.

firmware/rover-control-board/src/main.cpp
```diff
@@ -0,0 +1,304 @@
...
+    for (const auto& [name, id, power_bus] : BUS_GROUPS)
+    {
+        try
+        {
+            parameter_groups.emplace_back(power_bus.get_parameter_group());
+            packet_manager->add_group(parameter_groups.back());
+            packet_manager->set_transmission_config(
+                flagged_address_t(standard_address_t(RCB_DEVICE_ADDRESS, static_cast<uint8_t>(id), static_cast<uint8_t>(hi_can::addressing::power::distribution::rover_control_board::power_bus::parameter::POWER_STATUS))),
+                power_bus.get_transmission_config());
```
RandomSpaceship: If you use the parameter group's callback and transmission registration features, this code block should reduce to:
```suggestion
            packet_manager->add_group(power_bus.get_parameter_group());
```
There's also no need to copy the parameter group with the `parameter_groups` vector. Since it's part of the power bus class, the reference will never be invalidated as long as the backing power bus class is in scope.

firmware/rover-control-board/src/main.cpp
```diff
@@ -0,0 +1,304 @@
...
+                      CONFIG_AUX_PRECHARGE_VOLTAGE, RCB_AUX_PRE_SWITCH_PIN, RCB_AUX_MAIN_SWITCH_PIN,
+                      static_cast<gpio_num_t>(ROVER_PIN::A8), static_cast<gpio_num_t>(ROVER_PIN::A7),
+                      AUX_VOLTAGE_DIVIDER_HIGH_RESISTOR, AUX_VOLTAGE_DIVIDER_LOW_RESISTOR, AUX_CURRENT_SENSE_RESISTOR);
+
+const std::vector<std::tuple<std::string, power::distribution::rover_control_board::group, RoverPowerBus&>> BUS_GROUPS = {
+    {"compute", power::distribution::rover_control_board::group::COMPUTE_BUS, compute_bus},
+    {"drive", power::distribution::rover_control_board::group::DRIVE_BUS, drive_bus},
+    {"aux", power::distribution::rover_control_board::group::AUX_BUS, aux_bus},
+    {"spare", power::distribution::rover_control_board::group::SPARE_BUS, spare_bus},
+};
```
RandomSpaceship: You'd be better off with an ID -> name/bus map, then you can de-duplicate ID to name code further down and still iterate over it.

RandomSpaceship (commented):
Found a few issues. It's a fairly "faithful" port of the original, but that also means it comes with all the problematic semantics thereof, so it'll need some major work to bring it inline with code standards and the expected `hi_can` usage patterns. I'd strongly recommend **not** using/testing this code until _after_ the competition.

PR #410, opened 2026-03-12
Add Open3D 0.19.0 to our nix-managed environment

DingoOz:
This PR adds Open3D 0.19.0 as a Nix-managed package, built from source with Python 3.13 support. Open3D gives us point cloud processing, mesh operations, 3D registration (ICP), and visualisation. All   available out of the box in the dev shell.

- New Nix package at packages/open3d/ that builds Open3D from source in the Nix sandbox, pre-fetches some third-party dependencies that  can't use nixpkgs versions
- Uses system libraries where possible (eigen, fmt, embree, VTK, etc.) via USE_SYSTEM_* cmake flags
- Python module installs to site-packages with runtime deps (numpy,  plotly, dash) wired up via passthru.pythonDeps
- Added to all ROS workspaces through standardPkgs with PYTHONPATH set up in the shell hook
- Includes a plotly-based demo script at software/scripts/open3d_demo.py that runs in the browser

Please note:
 - x86_64-linux only. The package uses pre-built Intel MKL and IPP binaries, so meta.platforms is restricted accordingly
 - Have set BUILD_GUI=OFF . This is disabled to avoid additional dependencies; the legacy visualiser and plotly-based    visualisation both work fine though.

image

### DIFF
--- a/packages/open3d/default.nix
+++ b/packages/open3d/default.nix
@@ -26,6 +26,7 @@
   vtk,
   minizip,
   xorg,
+  librealsense,
   libGL,
   unzip,
 }:
@@ -135,6 +136,7 @@ stdenv.mkDerivation rec {
     xorg.libXinerama
     xorg.libXcursor
     xorg.libXi
+    librealsense
     libGL
   ];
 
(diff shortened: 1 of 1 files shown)


packages/open3d/default.nix
```diff
@@ -0,0 +1,280 @@
...
+        'GIT_REPOSITORY https://github.com/microsoft/DirectXMath.git' \
+        'URL file://${directxmath-src}' \
+      --replace-fail \
+        'GIT_TAG may2022' \
+        '# patched: tag removed'
+    # Remove GIT_SHALLOW (appears twice, once per git dep)
+    sed -i '/GIT_SHALLOW TRUE/d' 3rdparty/uvatlas/uvatlas.cmake
+  '';
+
+  cmakeFlags = [
```
BrockoliniMorgan: Should BUILD_LIBREALSENSE and USE_SYSTEM_LIBREALSENSE be added here (in addition to supplying the build with librealsense)? I'm not sure what is planned for the usage of open3d, but it might be good to have realsense support

UltraFishy (approved):
Brock brought up the only issues I saw. Looks good!

PR #411, opened 2026-03-13
Feat/add realsense urdf and gazebo 

bocho0600:
this PR replicate realsense d445 in simulation to match with the real camera (output topic + configuration)
image

Tested by running the simulation:

https://roar-qutrc.github.io/tutorials/running-simulation.html?h=simulation


software/ros_ws/src/perseus_description/urdf/camera_d455_gazebo.xacro
```diff
@@ -0,0 +1,105 @@
...
+            <width>640</width>
+            <height>480</height>
+            <format>R8G8B8</format>
+          </image>
+          <clip>
+            <near>0.05</near>
+            <far>100</far>
+          </clip>
+        </camera>
+        <topic>/camera/camera/color/image_raw</topic>
```
DingoOz: Gazebo's camera sensor type auto-appends /image and /camera_info to whatever you set in <topic>. Since the RGB topic is set to /camera/camera/color/image_raw, camera_info will actually publish at /camera/camera/color/image_raw/camera_info.  gz_bridge.yaml:47 expects it at /camera/camera/color/camera_info. Same issue applies to all sensor <topic> entries in this file (depth, infra1, infra2). Consider setting topics to the base path (e.g. /camera/camera/color) and letting Gazebo append the suffixes, then update the bridge image mappings to match.

DingoOz (changes requested):
perseus_simulation/rviz/view.rviz at line: 441 still subscribes to /rgbd_camera/points, which was the old Gazebo topic. Since you're remapping everything to the RealSense   topic convention, this should be /camera/camera/depth/color/points so the pointcloud displays correctly when launching the sim with RViz.

In perseus_simulation/rviz/view.rviz, multiple references to camera_link_optical in RViz display configs. This frame no longer exists so it needs to update to camera_color_optical_frame (or whichever optical frame is appropriate per display). Same in perseus_description/rviz/view_perseus.rviz

In urdf/sensors.urdf.xacro the Livox IMU sensor uses a relative topic scan_3d/imu. The bridge now correctly maps from absolute /scan_3d/imu (fixing the missing leading slash), but Gazebo may prepend the model namespace to relative sensor topics (e.g. /model/perseus/scan_3d/imu). Please verify this matches at runtime. If not, either make the sensor topic absolute in the xacro or adjust the bridge mapping.

software/ros_ws/src/perseus_description/urdf/camera_d455_gazebo.xacro
```diff
@@ -0,0 +1,105 @@
...
+        <gz_frame_id>${name}_color_optical_frame</gz_frame_id>
+      </sensor>
```
DingoOz: All the sensor topics are hardcoded to `/camera/camera/...` but the frame references use `${name}` so they adapt if the name parameter changes. If someone passes `name:=front_camera` the frames will update but the topics won't, so the bridge mappings will silently break. Could we parameterize the topic paths as `/${name}/${name}/color/image_raw` etc., or at minimum add a comment noting that topics are only correct when `name=camera`.

software/ros_ws/src/perseus_description/urdf/camera_d455_gazebo.xacro
```diff
@@ -0,0 +1,105 @@
+<?xml version="1.0"?>
+<robot xmlns:xacro="http://ros.org/wiki/xacro">
+
+  <xacro:macro name="camera_d455_gazebo" params="name:=camera">
+
+    <gazebo reference="${name}_color_frame">
+      <material>Gazebo/Red</material>
+      <sensor name="rgb_camera" type="camera">
```
DingoOz: There's both a standalone `camera` type sensor on `color_frame` and an `rgbd_camera` type sensor on `depth_frame`. The `rgbd_camera` already produces an RGB image (`/camera/camera/depth/image` shows up in your `gz topic -l` output), so this gives us two RGB streams in Gazebo which doubles the GPU work. Could we drop the standalone RGB camera and bridge the `rgbd_camera`'s image output instead? Or is there a reason we need both — e.g. different resolution or FOV for the color vs depth?

DingoOz (dismissed):
Thanks for addressing the earlier feedback, most of the items are resolved. A few things still need attention before this can merge.

**aruco_detector.hpp:81** — The `.cpp` default and YAML config were updated to `camera_color_optical_frame` but the header member initializer still says `camera_link_optical`. The `declare_parameter` call overwrites it at runtime so it works today, but if anyone reads this header for reference or removes the parameter declaration it will silently break ArUco TF lookups. Should be `camera_color_optical_frame` to match.

**perseus_description/rviz/view_perseus.rviz** — This file still references `camera_link_optical` (lines 78, 83, 181, 183, 225, 226). These frames no longer exist in the TF tree since the old `camera_depth.urdf.xacro` was removed. I flagged this in the first review — please update to the new RealSense frames the same way `perseus_simulation/rviz/view.rviz` was updated.

**perseus.urdf.xacro** — Missing trailing newline (`\ No newline at end of file`). This was fixed in `camera_d455_gazebo.xacro` but the same issue was introduced here.

software/ros_ws/src/perseus_description/urdf/camera_d455_gazebo.xacro
bocho0600: updated topic to take ${name}

software/ros_ws/src/perseus_description/urdf/camera_d455_gazebo.xacro
bocho0600: changeed camera type on depth to depth_camera (will not publish rgb) and set color to rgbd_camera to get aligned image

UltraFishy (approved):
Looking good 💯

PR #416, opened 2026-03-15
Date Fixed!

UltraFishy:
Added a daemon and install script for the daemon.
It is a startup daemon allowing for the `sudo date` to be created


LucaLubrano (approved):
looks good 👍

PR #418, opened 2026-03-15
Feat/lidarslam ros

BrockoliniMorgan:
Added all four ros2 packages from [Lidarslam-ros2](https://github.com/rsasaki0109/lidarslam_ros2) as well as one [external dependency](https://github.com/rsasaki0109/ndt_omp_ros2) for autonomy to use. I had to patch a couple of them because their package.xml files didn't include some dependencies, so the autogenerated files also didn't. I also patched ndt_omp_ros2 because it uses an outdated header from pcl which just included header files from boost. The descriptions of the patches are outlined in software/ros_ws/PATCHES.md
The lidarslam package depends on all of the other packages, so you can test by just adding lidarslam to the workspace.nix (or you can add all five if you want)

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -79,11 +79,11 @@
         "nixpkgs": "nixpkgs"
       },
       "locked": {
-        "lastModified": 1771881583,
-        "narHash": "sha256-vYTYv1PoGG/Q6afrKnj7fdT+h2f4b1auOdXmQZTaldU=",
+        "lastModified": 1773270149,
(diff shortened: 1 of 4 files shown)


DingoOz (changes requested):
Nice work getting these packaged up. The PATCHES.md documentation is thorough and the boost.h fix looks solid. I've flagged a few issues that need sorting before merge, mostly around build input categorisation in a couple of the packages and a docs mismatch.

PR #424, opened 2026-03-17
Feat: Implementing JSON lookup table for Hi-CAN addresses

test-account30:
## Hi-CAN Address JSON Generator

This tool parses `hi_can_address.hpp` and generates a flattened JSON lookup
table mapping CAN IDs to their corresponding system, subsystem, device,
group, and parameter names.

The goal is to provide a human-readable and machine-consumable reference for
debugging, logging and tooling that interacts with Hi-CAN messages.

---

## What this tool does

The generator:

- Reads the Hi-CAN address header file
- Parses namespace and enum definitions
- Computes full CAN IDs using the configured address bit layout
- Produces a JSON lookup table

Each entry maps a CAN ID to its semantic components.

To run the tool, invoke via command line:
```python3 hi-can-generator.py /path/to/hi_can_address.hpp /optional/path/to/output.JSON```

Example output:

```json
{
    "0x02000d03": {
    "system": "excavation",
    "subsystem": "bucket",
    "device": "controller",
    "group": "MAGNET",
    "parameter": "MAGNET_ENABLE"
  }
}

### CODE REVIEW

File: software/shared/hi-can-generator/hi-can-generator.py [hunk 1]
```diff
@@ -0,0 +1,244 @@
...
+            if "GROUP_ADDRESS_BITS" in token:
+                systems["GROUP_ADDRESS_BITS"] = token.split("GROUP_ADDRESS_BITS = ")[
+                    1
+                ].split(";")[0]
+            if "PARAM_ADDRESS_BITS" in token:
+                systems["PARAMETER_ADDRESS_BITS"] = token.split(
+                    "PARAM_ADDRESS_BITS = "
+                )[1].split(";")[0]
+
+    # Build flat lookup table
```
DingoOz: **Bug:** The last system before `legacy` is silently dropped. `systems.update(system)` only runs when the *next* `SYSTEM_ID` token is encountered, but when the loop hits `"legacy"` it breaks before reaching that branch. The `shared` system (0x06) and its elevator motor data will be missing from the JSON output.

Add `systems.update(system)` after the for loop, before building the flat lookup table (LUT):
```python
    # (end of for loop)
    systems.update(system)

    # Build flat lookup table
    flat_lut = {}
```

PR #427, opened 2026-03-18
Feat/add cube detector

bocho0600:
## Summary
This PR adds a full cube detection pipeline to `perseus_vision` with YOLOv8 + ONNX Runtime, including optional depth-based 3D localization, TF frame output, marker visualization, and ROS service/topic outputs for downstream autonomy.

### 1. 2D Cube Detection (RGB)
- New/updated `cube_detector` node performs cube detection from camera RGB images.
- Uses ONNX Runtime (YOLOv8 model) for inference.
- Applies confidence filtering + NMS.
- Publishes optional annotated image with:
  - bounding boxes,
  - class label/confidence,
  - center pixel text/crosshair.

### 2. Runtime Control for CPU Load
- `always_on` parameter controls continuous inference.
- `processing_frequency_hz` limits inference rate when running continuously.
- Supports low-load workflow:
  - keep `always_on=false`,
  - run detection only when needed via service call.

### 3. Depth-Based 3D Estimation (`depth_image` mode)
- Added `depth_estimation_mode` support (`none` / `depth_image`).
- Reads depth image + depth camera_info.
- Estimates cube 3D pose from depth at detection ROI center.
- Handles both `16UC1` and `32FC1` depth encodings.
- Applies range filtering (`min_range_m`, `max_range_m`).

### 4. TF Output Frame
- Added `tf_output_frame` parameter.
- Cube poses are transformed into the requested output frame before publishing.
- Works with RViz and downstream consumers expecting a specific frame (e.g. `odom`, `base_link`, `camera_link_optical`).

### 5. MarkerArray Output
- Publishes cube markers (`visualization_msgs/MarkerArray`) for RViz.
- Marker color matches detected cube color.
- Text marker shows cube `x,y` coordinates in output frame.

### 6. `ObjectDetections` Topic Output
- Publishes `perseus_interfaces/msg/ObjectDetections` for cube detections.
- Includes `stamp`, `frame_id`, `ids`, and `poses`.
- Cube color-to-ID mapping:
  - blue → 1
  - green → 2
  - red → 3
  - white → 4

### 7. `DetectObjects` Service Support
- `cube_detector` now serves `perseus_interfaces/srv/DetectObjects`.
- Returns latest detections (`stamp`, `frame_id`, `ids`, `poses`).
- Optional image capture:
  - saves annotated snapshot to requested path,
  - overlays timestamp, frame, bbox, and cube location metadata.

## Config/Usability Improvements
- Added configurable output topics:
  - annotated image topic,
  - detections topic,
  - markers topic.
- Improved model path robustness:
  - supports `~/...`,
  - handles legacy `/perseus-v2/...` style paths,
  - fallback handling if configured path is missing.

## How To Test

1. Enter dev environment
```bash
cd ~/perseus-v2
nix develop 
cd software/ros_ws
```

2. Build
```bash
colcon build --packages-select perseus_vision
source install/setup.bash
```

3. Launch cube detector
```bash
ros2 launch perseus_vision cube_detector.launch.py use_sim_time:=true
```

4. Verify topics
```bash
ros2 topic echo /perseus_vision/cube/detections --once
ros2 topic echo /perseus_vision/cube/markers --once
```

5. Service test (on-demand detection + image save)
```bash
ros2 param set /cube_detector always_on false
ros2 service call /detect_objects perseus_interfaces/srv/DetectObjects \
"{capture_image: true, img_save_path: '/tmp/cube_captures'}"
ls -lah /tmp/cube_captures
```

6. RViz check
- Add `MarkerArray` display on `/perseus_vision/cube/markers`
- Confirm markers and text appear in `tf_output_frame`.

7. Optional continuous mode check
```bash
ros2 param set /cube_detector always_on true
ros2 param set /cube_detector processing_frequency_hz 5.0
```
image
image


DingoOz (commented):
Overall this is a solid piece of work. The YOLOv8 + ONNX integration, depth estimation, and service/topic architecture are well thought out.

DingoOz (commented):
Nice work on this, the YOLOv8 + ONNX integration with depth estimation and service support is well thought out. Inline comments below.

software/ros_ws/src/perseus_vision/CMakeLists.txt
```diff
@@ -43,12 +52,42 @@ target_include_directories(
...
+  perseus_interfaces)
+
+# ── install ───────────────────────────────────────────────────────────────────
+install(TARGETS aruco_detector_node cube_detector
+        DESTINATION lib/${PROJECT_NAME})
+
+install(DIRECTORY include/ DESTINATION include/)
+install(DIRECTORY config/ DESTINATION share/${PROJECT_NAME}/config)
 install(DIRECTORY launch/ DESTINATION share/${PROJECT_NAME}/launch)
+install(DIRECTORY models/ DESTINATION share/${PROJECT_NAME}/models)
```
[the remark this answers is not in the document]
DingoOz: Model was removed from the repo in 7b3f5376, so this is resolved. Git LFS is still worth setting up before the next time a large binary needs to land.

software/ros_ws/src/perseus_vision/include/perseus_vision/cube_detector.hpp
```diff
@@ -0,0 +1,173 @@
...
+#include "sensor_msgs/msg/image.hpp"
+#include "std_msgs/msg/header.hpp"
+#include "std_msgs/msg/string.hpp"
+#include "tf2_ros/buffer.h"
+#include "tf2_ros/transform_listener.h"
+#include "visualization_msgs/msg/marker_array.hpp"
+namespace perseus_vision
+{
+
+    static const std::vector<std::string> CLASS_NAMES = {"blue", "green", "red", "white"};
```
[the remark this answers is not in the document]
DingoOz: Fixed in aa71e3cb. `CLASS_NAMES` and `CLASS_COLOURS` are now `inline const`. Good.

software/ros_ws/src/perseus_vision/src/cube_detector.cpp
```diff
@@ -0,0 +1,867 @@
...
+
+        auto detections = postprocess(out_data, num_boxes);
+        publish_annotated_image(bgr_image, detections, header);
+
+        // Refresh detection snapshot every inference call so service output is current.
+        {
+            std::lock_guard<std::mutex> lock(_detections_mutex);
+            _latest_detections_stamp = header.stamp;
+            _latest_detections_frame_id = _tf_output_frame;
+            _latest_detection_ids.clear();
```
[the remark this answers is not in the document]
DingoOz: Partially addressed in 1d1c74f3. The cached detections now only contain cubes with valid depth-derived poses, which is an improvement. But the original question still stands -- if depth estimation fails for every cube, the service will return zero results even though 2D detections were found. If the intent is that the service only returns depth-resolved cubes, that is fine, but it would be worth adding a log line or a note in the service response so callers know detections were found but depth was unavailable.

software/ros_ws/src/perseus_vision/src/cube_detector.cpp
```diff
@@ -0,0 +1,867 @@
...
+        : Node("cube_detector"),
+          _ort_env(ORT_LOGGING_LEVEL_WARNING, "cube_detector")
+    {
+        // ── parameters ──────────────────────────────────────────────────────────────
+        declare_parameter("model_path", "/perseus-v2/software/ros_ws/src/perseus_vision/models/cube_detector_yolob8s.onnx");
+        declare_parameter("confidence_threshold", 0.5);
+        declare_parameter("camera_topic", "/camera/camera/color/image_raw");
+        declare_parameter("camera_info_topic", "/camera/camera/color/camera_info");
+        declare_parameter("always_on", true);  // keep node alive even without subscribers
+        declare_parameter("use_cuda", true);
```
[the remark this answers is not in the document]
DingoOz: `use_cuda` fixed in adeb56a2, now defaults to `false` in both code and YAML. `confidence_threshold` still has a mismatch though -- code declares the default as 0.5 but YAML sets 0.3. Not critical since the YAML will win at runtime, but worth aligning so the node behaves sensibly if launched without config.

software/ros_ws/src/perseus_vision/src/cube_detector.cpp
```diff
@@ -0,0 +1,886 @@
...
+            output_names, 1);
+
+        const float* out_data = output_tensors[0].GetTensorData<float>();
+        const auto out_shape = output_tensors[0].GetTensorTypeAndShapeInfo().GetShape();
+        const size_t num_boxes = static_cast<size_t>(out_shape[2]);
+
+        auto detections = postprocess(out_data, num_boxes);
+        publish_annotated_image(bgr_image, detections, header);
+
+        if (_depth_estimation_mode == "depth_image")
```
DingoOz: `publish_depth_markers` is the only place that publishes to `_pub_cube_detections`. When `depth_estimation_mode` is `"none"`, this block is skipped and the detections topic never publishes anything. Downstream consumers subscribing to `/perseus_vision/cube/detections` will get nothing unless depth is enabled. The 2D detection results should be published regardless of depth mode.

software/ros_ws/src/perseus_vision/CMakeLists.txt
```diff
@@ -43,12 +52,42 @@ target_include_directories(
...
+ament_target_dependencies(
+  cube_detector
+  rclcpp
+  sensor_msgs
+  std_msgs
+  cv_bridge
+  ament_index_cpp
+  geometry_msgs
+  visualization_msgs
+  tf2_ros
```
DingoOz: `cube_detector.cpp` includes `tf2/exceptions.h` but `tf2` is not listed in `ament_target_dependencies` here. It works now via transitive dependency from `tf2_ros` but that is not guaranteed across ROS2 distros. The aruco_detector target already lists `tf2` explicitly, cube_detector should too.

bocho0600: added tf2 dependency to cube detector executable

software/ros_ws/src/perseus_vision/include/perseus_vision/cube_detector.hpp
```diff
@@ -0,0 +1,195 @@
...
+#include "opencv2/opencv.hpp"
+#include "perseus_interfaces/msg/object_detections.hpp"
+#include "perseus_interfaces/srv/detect_objects.hpp"
+#include "rcl_interfaces/msg/set_parameters_result.hpp"
+#include "rclcpp/rclcpp.hpp"
+#include "sensor_msgs/msg/camera_info.hpp"
+#include "sensor_msgs/msg/image.hpp"
+#include "sensor_msgs/msg/region_of_interest.hpp"
+#include "std_msgs/msg/header.hpp"
+#include "std_msgs/msg/string.hpp"
```
DingoOz: Small issue: This include appears unused, nothing in the class uses `std_msgs::msg::String`. Remove it?

software/ros_ws/src/perseus_vision/src/cube_detector.cpp
```diff
@@ -0,0 +1,980 @@
...
+        const std_msgs::msg::Header& header)
+    {
+        perseus_interfaces::msg::ObjectDetections detections_msg;
+        detections_msg.stamp = header.stamp;
+        detections_msg.frame_id = header.frame_id;
+
+        for (const auto& detection : detections)
+        {
+            geometry_msgs::msg::Pose unresolved_pose;
+            const double nan = std::numeric_limits<double>::quiet_NaN();
```
DingoOz: Publishing NaN-filled poses when depth isn't available could cause issues for downstream consumers that don't expect NaN (e.g., Nav2, TF operations, or anything that does arithmetic on the pose). Consider either:
- Omitting the `poses` array entirely in 2D-only mode, or
- Documenting this convention clearly in `ObjectDetections.msg` so consumers know to check for NaN.

software/ros_ws/src/perseus_vision/src/cube_detector.cpp
```diff
@@ -0,0 +1,980 @@
...
+        if (_model_path.rfind("~/", 0) == 0)
+        {
+            if (const char* home = std::getenv("HOME"))
+            {
+                _model_path = std::string(home) + _model_path.substr(1);
+            }
+        }
+
+        // Backward compatibility: some configs use "/perseus-v2/..." (missing $HOME prefix).
+        if (!_model_path.empty() && !std::filesystem::exists(_model_path))
```
DingoOz: The model path resolution chain is: expand `~/` → prepend `$HOME` → fall back to package share dir. A bare filename like `"cube_detector_yolob8s.onnx"` from the YAML config won't match any of these (since `std::filesystem::exists("cube_detector_yolob8s.onnx")` is CWD-dependent and will almost always fail), so it silently falls through to the share-dir fallback.

This works by accident but is fragile. Either handle relative paths explicitly (resolve against share dir), or document that `model_path` must be absolute or empty.

software/ros_ws/src/perseus_vision/include/perseus_vision/cube_detector.hpp
```diff
@@ -0,0 +1,195 @@
...
+#include "tf2_ros/buffer.h"
+#include "tf2_ros/transform_listener.h"
+#include "visualization_msgs/msg/marker_array.hpp"
+namespace perseus_vision
+{
+
+    inline const std::vector<std::string> CLASS_NAMES = {"blue", "green", "red", "white"};
+
+    inline const std::vector<cv::Scalar> CLASS_COLOURS = {
+        cv::Scalar(255, 100, 0),    // blue
```
DingoOz: Nit: `blue` is `Scalar(255, 100, 0)` in BGR — that's more of a teal/cyan than blue. Is this intentional for visibility on the annotated image, or should it be `Scalar(255, 0, 0)` for true blue?

DingoOz (commented):
Nice work on the cube detection pipeline, the ONNX integration, depth estimation, and dual topic/service interface are solid. A few things to address before merge, see inline comments.

software/ros_ws/src/perseus_vision/include/perseus_vision/cube_detector.hpp
bocho0600: image

this is the blue colour I am  using in the annotation, the reason is when I tested it with pure blue (255,0,0) I found the text and bounding box looks a bit dark to look at so I add a bit of green into it to make it a bit brighter but still looks blue

software/ros_ws/src/perseus_vision/src/cube_detector.cpp
bocho0600: model_path no longer relies on the accidental fallback path. If a relative filename is provided in the YAML, the node now explicitly resolves it against share/perseus_vision/models/ first, and only falls back to the default packaged model if that file is not found. I also updated the config comments to make the workflow clearer for custom models.

software/ros_ws/src/perseus_interfaces/srv/DetectObjects.srv
```diff
@@ -17,3 +17,7 @@ int32[] ids
 # Pose of each detected marker, in the same order as `ids`
 # (ids[i] corresponds to poses[i])
 geometry_msgs/Pose[] poses
+
+# Human-readable status for callers, e.g. when 2D detections were found but
+# could not be resolved into poses.
+string message
```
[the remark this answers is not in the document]
bocho0600: added regions_of_interest to the service

DingoOz (approved):
Looks good. Well-structured cube detection pipeline, all previous feedback addressed across multiple review rounds, CI green. Approve.

PR #430, opened 2026-03-25
feat: Added CAN-debug UI

ZephCreations:
Added Web-UI component for Can messages. It will display the most recent data from each can message. 
It works by reading the output of "candump" and parsing it to a buffer. It then sends the data to the svelte front-end every two seconds which updates with the latest data. 

No messages:
image

With messages:
image


software/web_ui/src/server/scripts/canSocket.js
```diff
@@ -0,0 +1,144 @@
+// @ts-nocheck
+import { spawn } from "child_process";
+import path from "path";
+import fs from "node:fs";
```
DingoOz, line 4: Unused import. `response` from `express` is never used in this file. Remove it.

software/web_ui/src/server/scripts/canSocket.js
```diff
@@ -0,0 +1,144 @@
...
+const canLookupFilePath = path.resolve("src/lib/canLookup.json");
+
+let timeoutId;
+let running = false;
+
+let canLookup = null;
+const buffer = [];
+const MAX_BUFFER = 1000;
+
+export const canSocket = (socket) => {
```
DingoOz: **Bug:** This function receives `io` (the Socket.IO server instance) from both `server.js` and `vite.config.ts`, but the parameter is named `socket` and used as if it's a per-client socket. `io.emit()` happens to broadcast to all clients which may accidentally work, but the semantics are wrong — if you intended per-client delivery you need the actual socket, and if you intended broadcast you should name the parameter `io` for clarity.

Either rename the parameter to `io`, or change the call sites to pass it inside `io.on("connection")` like `cameraSocket` does.

software/web_ui/src/server/scripts/canSocket.js
```diff
@@ -0,0 +1,144 @@
...
+  if (!running) {
+    // Create lookup table file on launch
+    generateFile();
+
+    // Start processing candump output
+    const stopCandump = startCanDump("can0");
```
DingoOz: `stopCandump` is assigned but never used. The `candump` child process will run forever with no cleanup — no disconnect handler, no shutdown hook. The `setInterval` stored in `timeoutId` is also never cleared.

At minimum, store `stopCandump` at module level and add a cleanup path. Same issue exists in `resourceMonitorSocket` but no need to inherit that pattern.

software/web_ui/src/server/scripts/canSocket.js
```diff
@@ -0,0 +1,144 @@
...
+    });
+  });
+};
+
+const getCanData = () => {
+  return new Promise((resolve, reject) => {
+    const temp = latestInfo;
+    latestInfo = [];
+    resolve(latestInfo);
+  });
```
DingoOz: **Dead code:** `getCanData()` references `latestInfo` which is undefined. It's also never called. Remove it.

software/web_ui/src/server/scripts/canSocket.js
DingoOz: **Race condition:** `generateFile()` is async (spawns python, awaits the result) but `startCanDump("can0")` is called immediately after without awaiting. If CAN frames arrive before the python process finishes, `canLookup` is still `null` and `parseCandump` will set `details: undefined` on every parsed frame.

Either `await generateFile()` before starting candump, or add a null check in `parseCandump` to skip frames when the lookup table isn't ready yet.

software/web_ui/src/lib/widgets/canDebug.svelte
```diff
@@ -0,0 +1,124 @@
...
+		dataList.forEach((value, index) => {
+			const inArray = canMsgs.findIndex((msg) => msg.address === value.address);
+
+			if (inArray === -1) {
+				// Add new row
+				canMsgs.push({
+					address: value.address,
+					system: value.details.system,
+					subsystem: value.details.subsystem,
+					device: value.details.device,
```
DingoOz: **Bug:** This uses `index` (the iteration index from `dataList`) instead of `inArray` (the found index in `canMsgs`). If the incoming data order doesn't match the order in `canMsgs`, this updates the wrong row.
```js
// should be:
canMsgs[inArray].latestData = value.data;
```

software/web_ui/src/lib/widgets/canDebug.svelte
```diff
@@ -0,0 +1,124 @@
...
+		{
+			address: string;
+			system: string;
+			subsystem: string;
+			device: string;
+			group: string;
+			parameter: string | null;
+			latestData: number;
+		}[]
+	>([]);
```
DingoOz, line 31: Type mismatch — `latestData` is typed as `number` here, but the server sends `data: parts.slice(3)` which is `string[]`. Either update the type to `string[]` or parse the data to a number on the server side.

software/web_ui/src/lib/canLookup.json
```diff
@@ -0,0 +1,303 @@
+{
```
DingoOz: This is a 303-line generated file that's also regenerated at runtime by `generateFile()`. Should it be committed? If the runtime generation works, this file will go stale every time the hi-can address header changes. Either treat it as generated (gitignore it) or remove the runtime generation and use it as the static source of truth.

DingoOz (changes requested):
Good feature — having CAN bus visibility in the web UI is useful. The Svelte widget is clean and follows the existing widget pattern well. There are a few bugs in the server-side socket script that need fixing before merge though. See inline comments.

software/web_ui/src/server/scripts/canSocket.js
ZephCreations: Removed unused import (dunno how that got in there)

software/web_ui/src/server/scripts/canSocket.js
ZephCreations: Fixed to broadcast as it is a lightweight widget (any filtering or ordering will be client-side as a later UI improvement)

software/web_ui/src/server/scripts/canSocket.js
ZephCreations: Removed code

software/web_ui/src/lib/widgets/canDebug.svelte
ZephCreations: Fixed

software/web_ui/src/server/scripts/canSocket.js
ZephCreations: Added Await and null check (now handles incorrect can data by ignoring, can be fixed later but isn't strictly needed)

software/web_ui/src/lib/widgets/canDebug.svelte
ZephCreations: Changed type to string[]. No numerical processing is happening so string value is fine.

software/web_ui/src/lib/canLookup.json
ZephCreations: Added to .gitignore, it is generated on startup to reference and ensure it is always up to date.

software/web_ui/src/server/scripts/canSocket.js
ZephCreations: Added cleanup path and cleared command call.

software/web_ui/src/server/scripts/canSocket.js
```diff
@@ -0,0 +1,179 @@
...
+
+    process.on("SIGTERM", () => {
+      cleanup();
+      process.exit(0);
+    });
+  }
+}
+
+function cleanup() {
+  if (cleaned) return;
```
DingoOz, line 56: **Bug:** `cleaned` is never declared. This will throw a `ReferenceError` when `SIGINT`/`SIGTERM` fires, meaning the cleanup never actually runs. The `candump` child process leaks and the interval keeps ticking.

Add `let cleaned = false;` up with the other module-level state (near line 16).

software/web_ui/src/server/scripts/canSocket.js
```diff
@@ -0,0 +1,179 @@
...
+    });
+
+    pythonProcess.on("close", (code) => {
+      if (code !== 0) {
+        reject(error);
+        return;
+      }
+
+      try {
+        canLookup = JSON.parse(output);
```
DingoOz: `canLookup` is assigned here inside `lookupTable()`, and then assigned again in `generateFile()` on line 69. The double assignment is harmless, but this also parses the JSON twice for no reason. Parse once:
```js
const data = JSON.parse(output);
canLookup = data;
resolve(data);
```
Then remove the redundant `canLookup = data;` in `generateFile()`. Let one function own the assignment.

software/web_ui/src/server/scripts/canSocket.js
```diff
@@ -0,0 +1,179 @@
...
+let stopCandump = null;
+
+let canLookup = null;
+const buffer = [];
+const MAX_BUFFER = 1000;
+
+export async function canSocket(io) {
+  if (!running) {
+    // Create lookup table file on launch
+    await generateFile();
```
DingoOz, line 28: `generateFile()` has a `try/catch` (line 71) that swallows errors. If the python process fails (missing file, python3 not installed, invalid JSON), the error is logged but `canLookup` stays `null`. Execution continues into `startCanDump` as if everything is fine.

The first CAN frame that arrives then hits `canLookup[address]` on line 155, which throws `TypeError: Cannot read properties of null`. That's an uncaught exception inside a stream `data` handler, which crashes the entire web server.

Either re-throw from `generateFile()` so `canSocket()` can bail out, or add a null guard in `parseCandump`.

software/web_ui/src/server/scripts/canSocket.js
```diff
@@ -0,0 +1,179 @@
...
+      } catch (err) {
+        console.error(err);
+        reject(`Invalid JSON\n`);
+      }
+    });
+  });
+};
+
+export function startCanDump(iface) {
+  const proc = spawn("candump", [iface]);
```
DingoOz, line 114: No `error` event handler on the `candump` child process. If `candump` isn't installed or `can0` doesn't exist, Node emits an unhandled `error` event which crashes the entire web server, not just this widget.
```js
proc.on('error', (err) => {
  console.error('Failed to start candump:', err.message);
});
```

software/web_ui/src/server/scripts/canSocket.js
```diff
@@ -0,0 +1,179 @@
...
+
+    const lines = leftover.split("\n");
+    leftover = lines.pop() || "";
+
+    for (const line of lines) {
+      if (!line.trim()) continue;
+
+      const parsed = parseCandump(line);
+
+      if (!parsed.details || !parsed.data.length) {
```
DingoOz: **Bug:** `parseCandump` returns `null` when `parts.length < 4` (line 152), but this line accesses `parsed.details` before the null check on line 131. A malformed CAN line will throw `TypeError` here. The null guard needs to come first:
```js
const parsed = parseCandump(line);
if (!parsed) continue;

if (!parsed.details || !parsed.data.length) {
  continue;
}

buffer.push(parsed);
```

software/web_ui/src/server/scripts/canSocket.js
```diff
@@ -0,0 +1,179 @@
...
+  return parsed;
+}
+
+function formatTimestamp(ts) {
+  const d = new Date(ts);
+
+  const hours = String(d.getHours()).padStart(2, "0");
+  const minutes = String(d.getMinutes()).padStart(2, "0");
+  const seconds = String(d.getSeconds()).padStart(2, "0");
+  const ms = String(d.getMilliseconds()).padStart(2, "0");
```
DingoOz: Milliseconds are 0-999 (3 digits), but this pads to 2. Timestamps like `12:34:56.05` will appear for 5ms instead of `12:34:56.005`.
```js
const ms = String(d.getMilliseconds()).padStart(3, "0");
```

software/web_ui/src/lib/widgets/canDebug.svelte
```diff
@@ -0,0 +1,132 @@
...
+	// full lookup table
+	interface CanData {
+		iface: string;
+		address: string;
+		timestamp: string;
+		details: CanLutEntry;
+		data: string[];
+	}
+
+	const socket: Socket = io();
```
DingoOz: `const socket: Socket = io()` runs at the top level of the component script, outside `onMount`. The connection opens during component initialization, before the DOM is mounted. If this component is rendered during SSR or conditionally created and destroyed before mount, the `onMount` cleanup (line 82) never runs and the socket leaks.

Move the socket creation and event listener setup inside `onMount`.

DingoOz (changes requested):
Most of the issues from the first review are fixed, nice work. A couple of new bugs crept in with the cleanup and timestamp additions, and there are a few remaining issues in the error handling paths. See inline comments.

software/web_ui/src/server/scripts/canSocket.js
DingoOz: @OMEN44 you might also want to review this one directly.

software/web_ui/src/server/scripts/canSocket.js
ZephCreations: Added module-level definition

software/web_ui/src/server/scripts/canSocket.js
ZephCreations: Removed double reference. Assignment is now in Generate file as lookupTable() should just return the JSON

software/web_ui/src/server/scripts/canSocket.js
ZephCreations: Added null check before (into first parsed details if block, as if null check fails, others will not run.

software/web_ui/src/lib/widgets/canDebug.svelte
ZephCreations: Moved.

software/web_ui/src/server/scripts/canSocket.js
ZephCreations: Fixed to 3 decimals

software/web_ui/src/server/scripts/canSocket.js
ZephCreations: Added null check to parseCandump. Throwing could prevent webserver from running.

software/web_ui/src/server/scripts/canSocket.js
ZephCreations: Fixed error andling, it will correctlly handle the error and provide appropriate messages. Also performs cleanup to avoid any leaks.

DingoOz:
Hey @ZephCreations just checking in on this one. It looks like you addressed most of the review feedback back in March, nice work. A couple of items I don't see replies for:▎
 
1. Missing error handler on the candump child process (canSocket.js  ~line 114). If candump isn't installed or the interface doesn't exist, this will crash the whole web server.
 
2. Socket creation outside onMount (canDebug.svelte). The socket should be created inside onMount to avoid leaks during SSR or conditional rendering.
 
Could you check whether those were included in your fixes? If so, happy to do a final re-review. If this is still something you're working on, let me know and we can figure out next steps.

PR #433, opened 2026-04-23
feat: Added watch_dir to envrc to reload env

BrockoliniMorgan:
Just added watch_dir to the envrc so if any files in nix/ or software/ros_ws/nix-packages/ are changed the environment is reloaded (already happens with the `flake.nix` and `flake.lock` automatically). Prevents having to run `direnv reload` whenever running the `nix-package.sh` script. I think we *can* also use `watch_file *.nix` to reload whenever any nix file is changed, but this seems overkill and may become annoying. Thoughts on this are welcome.

Test using `touch nix/nixpkgs-config.nix` or similar and see the environment reload.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -4,6 +4,9 @@ if ! has nix_direnv_version || ! nix_direnv_version 3.0.7; then
   source_url "https://raw.githubusercontent.com/nix-community/nix-direnv/3.0.7/direnvrc" "sha256-bn8WANE5a91RusFmRI7kS751ApelG02nMcwRekC/qzc=" &>/dev/null
 fi
 
+watch_dir nix/
+watch_dir software/ros_ws/nix-packages/
+
 use flake .
 # print out autocomplete instructions
 printf '\033[0;33m' # orange text (makes it stand out more)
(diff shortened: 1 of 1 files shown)


DingoOz:
Nice fix. One thing worth considering is that this probably only covers less than half of the .nix files the flake actually evaluates, so editing things like software/ros_ws/overlay.nix, software/ros_ws/patches.nix, software/ros_ws/third-party-packages/**, software/shared/**, packages/**, software/overlay.nix, or treefmt.nix will still silently leave a stale dev shell.
 
Could the watch set be expanded to include?:
watch_dir nix/
watch_dir software/ros_ws/nix-packages/
watch_dir software/ros_ws/third-party-packages/
watch_dir software/shared/
watch_dir packages/
watch_file treefmt.nix software/overlay.nix software/ros_ws/overlay.nix software/ros_ws/patches.nix

Reloads are cached by nix_direnv, so the extra watches should not be costly. A one-line comment in .envrc explaining why these dirs are watched might also help future readers.

BrockoliniMorgan:
We could use watch_file to watch all .nix files with a glob if that would be preferable?

PR #436, opened 2026-04-25
Feat: Arm ROS2 software and hi_can definitions

OMEN44:
ARCh 26 Robot Arm ROS2 code.

This contains the ROS2 driver for each motor type and the arm controller node which handles communication between the two drivers and the control interface (web-ui)

This also contains the hi-can addresses and parameters for the arm (The RMD related hi-can definitions are strange and do not follow hi-can as they have been designed to fit the spec of the CAN hardware)


software/shared/hi-can/include/hi_can_address.hpp
```diff
@@ -365,6 +365,86 @@ namespace hi_can
...
+                    struct servo_address_t : public flagged_address_t
+                    {
+                        constexpr servo_address_t() = default;
+                        constexpr servo_address_t(const message_type function, const motor_id_t motor_id = motor_id_t::ALL)
+                            : flagged_address_t(static_cast<uint16_t>(function) + static_cast<uint8_t>(motor_id), false, false, false)
+                        {
+                        }
+                        constexpr uint8_t get_motor_id()
+                        {
+                            return (address && 0x000F);
```
DingoOz, line 407: This looks like it should be a bitwise AND. `address && 0x000F` will always evaluate to 0 or 1 regardless of the address value. Should be `(address & 0x000F)`.

software/ros_ws/src/perseus_payloads/src/arm_driver/rmd_driver.cpp
```diff
@@ -0,0 +1,276 @@
...
+}
+
+void RmdDriver::_handle_position_control(actuator_msgs::msg::Actuators servo_control)
+{
+    // RCLCPP_INFO(this->get_logger(), "Received position control command for RMD servos");
+    for (const auto& [motor_id, param_group] : this->PARAMETER_GROUP_MAP)
+    {
+        // For each online servo, send position command
+        const uint8_t index = static_cast<uint8_t>(motor_id) - 1;
+        const double target_position = servo_control.position[index];
```
DingoOz, line 95: There is no bounds check on `servo_control.position` or `servo_control.velocity` before indexing. If a publisher sends an Actuators message with fewer entries than the highest motor_id index, this will be undefined behaviour. Worth checking `position.size()` and `velocity.size()` against the indexes used before iterating.

software/ros_ws/src/perseus_payloads/src/arm_driver/rmd_driver.cpp
```diff
@@ -0,0 +1,276 @@
...
+    // Add RMD parameter groups to packet manager
+    for (const auto& [motor_id, parameter_group] : this->PARAMETER_GROUP_MAP)
+    {
+        _packet_manager->add_group(*parameter_group);
+    }
+    _packet_timer = this->create_wall_timer(PACKET_HANDLE_MS, std::bind(&RmdDriver::_handle_can, this));
+
+    // Initialise all motors on startup with desired functions
+    // Stop all motors - motors will lock up and won't flop around
+    _can_interface->transmit(Packet{
```
DingoOz, line 31: If the CAN init catch block above was hit, `_can_interface` is an empty optional and this `transmit` call will throw `std::bad_optional_access` straight back out of the constructor. The log line in the catch suggests we want to keep going on init failure, but as written we still take the node down. Either rethrow after logging or guard the rest of the constructor on init success.

software/ros_ws/src/perseus_payloads/src/arm_driver/rmd_driver.cpp
```diff
@@ -0,0 +1,276 @@
...
+        servo_address_t(message_type::SEND, current_id),
+        send_message::function_control_t(function::SET_CANID, static_cast<uint32_t>(request->data)).serialize_data()});
+    std::this_thread::sleep_for(PACKET_DELAY_MS);
+    // Since the ID is currently in an undefined state, MULTI_MOTOR_SEND must be used to ensure the motor is reset
+    _can_interface->transmit(Packet{
+        servo_address_t(message_type::MULTI_MOTOR_SEND),
+        send_message::action_command_t(send_message::action_command_t::command_id_t::RESET).serialize_data()});
+    std::this_thread::sleep_for(PACKET_DELAY_MS);
+
+    PARAMETER_GROUP_MAP.at(current_id)->set_online(false);  // mark old ID as offline
```
DingoOz, line 204: `.at()` throws `std::out_of_range` if `current_id` is not in `PARAMETER_GROUP_MAP`, which is the expected case when provisioning a fresh motor that still has its factory ID. Since this is a service callback, the exception will propagate out of the handler. Worth using `find()` and checking the iterator first.

software/ros_ws/src/perseus_payloads/src/arm_driver/end_effector.cpp
```diff
@@ -0,0 +1,105 @@
...
+        DEVICE_ID};
+    address.parameter = static_cast<uint8_t>(pwm_parameters::SET_PWM);
+    pwm_t pwm_command{};
+
+    RCLCPP_INFO(this->get_logger(), "Received write command with %zu data bytes", msg->data.size());
+
+    uint8_t index = 0;
+    for (const auto& group : {pwm_group::PWM_1, pwm_group::PWM_2, pwm_group::PWM_3, pwm_group::PWM_4, pwm_group::PWM_5})
+    {
+        pwm_command.value = msg->data[index++];  // Expecting a single uint16 value for PWM command
```
DingoOz, line 58: No size check on `msg->data` before this loop indexes into it five times. If a publisher sends fewer than 5 entries, which seems likely during a partial gripper update, this is UB. Worth a size check against the expected count before the loop.

software/ros_ws/src/perseus_payloads/include/arm_controller/arm_controller.hpp
```diff
@@ -0,0 +1,53 @@
...
+//          /positions (subscriber Float64MultiArray) - Current positions of all servos
+//     /rmd
+//         /control (publisher Actuators) - target positions for servos
+//         /status (subscriber Float64MultiArray) - status messages from servos
+//     /control (subscriber Float64MultiArray) - High level control parameters for the arm
+//     /position (publisher Float64MultiArray) - High level control parameters for the arm
+
+#define STATUS_FIELD_COUNT 8
+
+class ArmController : public rclcpp::Node
```
DingoOz, line 21: This header is missing `#pragma once`. `end_effector.hpp` and `rmd_driver.hpp` are also missing it. `rsbl_driver.hpp` already has it, and the codebase standard in CLAUDE.md mandates `#pragma once` on all headers, so the omission here looks unintentional.

software/ros_ws/src/perseus_payloads/include/arm_driver/rsbl_driver.hpp
```diff
@@ -0,0 +1,72 @@
...
+             std::make_shared<hi_can::parameters::post_landing::arm::control_board::rsblParameterGroup>(
+                 hi_can::addressing::post_landing::arm::control_board::rsbl_group::SHOULDER_PAN)},
+            {hi_can::addressing::post_landing::arm::control_board::rsbl_group::SHOULDER_TILT,
+             std::make_shared<hi_can::parameters::post_landing::arm::control_board::rsblParameterGroup>(
+                 hi_can::addressing::post_landing::arm::control_board::rsbl_group::SHOULDER_TILT)},
+        };
+
+    // Motor feedback
+    constexpr static auto POSITION_PUBLISH_MS = std::chrono::milliseconds(100);
+    constexpr static auto STATUS_REQUEST_MS = std::chrono::milliseconds(500);
```
DingoOz, line 62: `STATUS_REQUEST_MS` is declared but I cannot find it referenced anywhere in `rsbl_driver.cpp`. RmdDriver actively enables periodic status replies from the motors in its constructor, but I do not see the equivalent for RSBL. As-is, `/arm/rsbl/status` looks like it will just publish whatever defaults are sitting in `_status_1`/`_status_2`.

software/ros_ws/src/perseus_payloads/src/arm_driver/rmd_driver.cpp
```diff
@@ -0,0 +1,276 @@
...
+    }
+
+    _status_publisher->publish(status_msg);
+}
+
+void RmdDriver::_publish_motor_positions()
+{
+    std_msgs::msg::Float64MultiArray position_msg;
+
+    position_msg.data.resize(3, 0.0);
```
DingoOz, line 133: The resize to 3 elements here, combined with `_publish_motor_positions` only writing to indices [0] and [1], means slot [2] is always zero on this topic. Over in arm_controller, `_receive_rmd_positions` copies all three slots across and `_receive_rsbl_positions` then writes [2] with shoulder_tilt. It works today because of the single-threaded executor, but a callback-group or executor change would race on `_current_arm_positions[2]`. Probably cleaner to size by the number of RMD motors and have arm_controller write the RSBL slots at known offsets.

software/ros_ws/src/perseus_payloads/src/arm_driver/rmd_driver.cpp
```diff
@@ -0,0 +1,276 @@
...
+        _packet_manager->add_group(*parameter_group);
+    }
+    _packet_timer = this->create_wall_timer(PACKET_HANDLE_MS, std::bind(&RmdDriver::_handle_can, this));
+
+    // Initialise all motors on startup with desired functions
+    // Stop all motors - motors will lock up and won't flop around
+    _can_interface->transmit(Packet{
+        servo_address_t{message_type::MULTI_MOTOR_SEND},
+        send_message::action_command_t(send_message::action_command_t::command_id_t::STOP).serialize_data()});
+    std::this_thread::sleep_for(PACKET_DELAY_MS);
```
DingoOz, line 34: This sleep, plus three more in `_enable_status_messages`, means the constructor blocks the executor thread for around 125 ms on every node start. Could be deferred to a one-shot timer or moved into an `on_configure` lifecycle transition so the node finishes constructing first.

software/ros_ws/src/perseus_payloads/src/arm_driver/rmd_driver.cpp
```diff
@@ -0,0 +1,276 @@
...
+void RmdDriver::_set_motor_id(const std::shared_ptr<perseus_interfaces::srv::TriggerDevice::Request> request, std::shared_ptr<perseus_interfaces::srv::TriggerDevice::Response> response)
+{
+    using function = hi_can::parameters::post_landing::arm::rmd_servo::send_message::function_control_t::function_index_t;
+
+    const motor_id_t current_id = static_cast<motor_id_t>(request->id);
+
+    _can_interface->transmit(Packet{
+        servo_address_t(message_type::SEND, current_id),
+        send_message::function_control_t(function::SET_CANID, static_cast<uint32_t>(request->data)).serialize_data()});
+    std::this_thread::sleep_for(PACKET_DELAY_MS);
```
DingoOz, line 197: `sleep_for` inside a service callback stalls the executor for the duration of the sleep, plus another 25 ms below, plus whatever `_enable_status_messages` adds. Other timers and subscribers will be blocked for that whole window. Worth running this on a separate callback group, or breaking the sequence into a state machine driven by acknowledgements.

software/ros_ws/src/perseus_payloads/src/arm_driver/rsbl_driver.cpp
```diff
@@ -0,0 +1,158 @@
...
+    if (msg->position.size() >= 3)
+    {
+        using namespace hi_can::addressing;
+
+        position_control_t position_control{};
+
+        for (const auto& [servo_id, parameter_group] : this->PARAMETER_GROUP_MAP)
+        {
+            const int ID = static_cast<uint8_t>(servo_id);
+            position_control.position = static_cast<int16_t>(msg->position[ID - 1]);
```
DingoOz, line 62: `static_cast<int16_t>` will silently wrap if a publisher sends an out-of-range double. Combined with there being no joint-limit clamping anywhere upstream in arm_controller, the topic interface into the arm has no software bounds at all. Worth at least clamping to the joint's documented range before this cast.

software/ros_ws/src/perseus_payloads/src/arm_controller/arm_controller.cpp
```diff
@@ -0,0 +1,185 @@
...
+    //     double velocity = (position_diff / target_ms) * 1000.0;  // in degrees per second
+    //     velocities.push_back(velocity);
+    // }
+
+    using namespace hi_can::addressing::post_landing::arm::rmd_servo;
+
+    const uint8_t WRIST_PAN_ID = static_cast<size_t>(motor_id_t::WRIST_PAN) - 1;  // index starts at zero but id's start at 1
+    const uint8_t WRIST_TILT_ID = static_cast<size_t>(motor_id_t::WRIST_TILT) - 1;
+
+    double wrist_pos_a = msg->data[WRIST_PAN_ID] + msg->data[WRIST_TILT_ID];
```
DingoOz, line 142: This differential mixing uses unit coefficients on both pan and tilt. For most differential wrist mechanisms there is a coupling ratio or a factor of 1/2 so that pure pan and pure tilt produce equal-amplitude joint moves. Worth confirming with the mechanical team that the 1:1 sum-and-difference is what's actually wanted, otherwise tilt commands will translate to roughly double the intended tilt motion.

software/ros_ws/src/perseus_payloads/include/arm_driver/rmd_driver.hpp
```diff
@@ -0,0 +1,84 @@
...
+        const std::shared_ptr<perseus_interfaces::srv::TriggerDevice::Request> request,
+        std::shared_ptr<perseus_interfaces::srv::TriggerDevice::Response> response);
+    void _restart_motor(
+        const std::shared_ptr<perseus_interfaces::srv::TriggerDevice::Request> request,
+        std::shared_ptr<perseus_interfaces::srv::TriggerDevice::Response> response);
+    void _get_rmd_can_ids(
+        const std::shared_ptr<perseus_interfaces::srv::RequestInt8Array::Request> request,
+        std::shared_ptr<perseus_interfaces::srv::RequestInt8Array::Response> response);
+
+    constexpr static uint16_t RMD_SPEED_LIMIT = UINT16_MAX;  // max speed
```
DingoOz, line 51: `RMD_SPEED_LIMIT = UINT16_MAX` does not really line up with the hardware. The wire field is 1 dps per LSB per the RMD spec, so 65535 dps is well past any motor's mechanical limit. It also does not appear to be referenced anywhere. Either tie this to the actual joint limit in degrees per second, or remove it?

PR #441, opened 2026-06-02
Feat/taranis controller

GR3Y-SCALE:
Taranis controller has been added with full control and tweaked velocity values to give best performance. Generic controller has also been tweaked to change deadman functionality, namely stopping the turbo button acting as a deadman switch, so both bumpers will need to be depressed to make it go fast, instead of only hitting turbo button


software/ros_ws/src/perseus_input/src/generic_controller/generic_controller.cpp
```diff
@@ -174,8 +174,8 @@ GenericController::AxisParser::AxisParser(GenericController& parent, const std::
 double GenericController::AxisParser::get_value()
 {
     if (_has_enable &&
-        !_parent._enable_parsers.at(_param_base_name + ".enable").get_value() &&
-        !_parent._enable_parsers.at(_param_base_name + ".turbo_enable").get_value())
+        !_parent._enable_parsers.at(_param_base_name + ".enable").get_value())
+        //!_parent._enable_parsers.at(_param_base_name + ".turbo_enable").get_value())
```
DingoOz, line 178: Please delete this commented-out line rather than keeping it. As written we have a brace-less `if` with a dead comment sitting between the condition and the `return` - exactly the construct where a future edit (someone uncommenting this, or adding a statement) silently changes the logic. Suggest removing the comment and adding braces:

```cpp
if (_has_enable &&
    !_parent._enable_parsers.at(_param_base_name + ".enable").get_value())
{
    return 0.0;
}
```

software/ros_ws/src/perseus_input_config/config/taranis_controller.yaml
```diff
@@ -0,0 +1,38 @@
+generic_controller:
+  ros__parameters:
+    max_update_rate: 20
+    turbo_scaling: 2.0
+    drive:
+      forward:
+        turbo: -2.5
```
DingoOz, line 7: `turbo: -2.5` commands up to 2.5 m/s on `joy_vel`, but `diff_drive_base_controller` clamps `linear.x.max_velocity` at 1.7 m/s, so roughly the top 30% of the turbo stick range does nothing. Not unsafe (it just saturates), but if the goal was "tweaked velocity values for best performance", consider `-1.7` here, or add a comment noting the saturation is intentional.

software/ros_ws/src/perseus_input_config/config/taranis_controller.yaml
```diff
@@ -0,0 +1,38 @@
...
+        axis: 0
+        scaling: -0.7
+        enable:
+          axis: 4
```
DingoOz, line 11: This is the safety-critical channel - see my main comment about verifying the receiver's signal-loss failsafe puts this axis on the disarmed side of the -0.2 threshold. Worth recording the failsafe switch position in a comment here once tested.

software/ros_ws/src/perseus_input_config/config/taranis_controller.yaml
```diff
@@ -0,0 +1,38 @@
...
+      lift:
+        scaling: -0.8
+        axis: 3
+      tilt:
+        scaling: 0.8
+        axis: 1
+      jaws:
+        scaling: 0.1
+        axis: 6
+        #rotate:
```
DingoOz, line 36: Nit: either configure `rotate`/`magnet` or drop these commented keys - the parsers default safely to no-axis anyway, so the comments just add noise.

firmware/excavation-bucket/.python-version
```diff
@@ -0,0 +1 @@
+3.13.0
```
DingoOz, line 1: This file (and the matching one in `firmware/light-tower/`) looks unrelated to the Taranis work - local pyenv/uv artifacts? Suggest dropping both from this PR, and maybe adding `.python-version` to `.gitignore` if they keep appearing.

DingoOz (commented):
Took a close look at this since it changes deadman behaviour for every controller, not just the Taranis.

**The deadman logic change checks out.** I traced `AxisParser::get_value()`: previously holding the turbo trigger alone both armed the drive axes *and* applied 2.5x scaling, so there was a single-input path to max speed. With this change `enable` is always required to arm, and `turbo_enable` only switches the scaling (the turbo branch further down already checks it independently). Two inputs for turbo is strictly safer, and the Taranis config signs are consistent (turbo replaces scaling, and the signs match per axis).

A few things before this merges:

1. **This affects all controllers.** Xbox/Logitech/8bitdo drivers will find the turbo trigger alone no longer moves the rover. That's the intent, but the PR title doesn't flag it - anyone bisecting "turbo stopped working" will land on a commit about the Taranis. Worth a title tweak or a separate commit for the generic_controller change, plus a heads-up to whoever drives regularly.
2. **RC failsafe needs a hardware check.** The 150ms joy timeout only fires if `/joy` stops publishing, but RC receivers in USB-gamepad mode typically keep publishing failsafe axis values on signal loss. Please verify that with the transmitter off (or out of range) the enable switch channel (axis 4) lands on the disarmed side of the -0.2 threshold - otherwise signal loss with sticks deflected keeps the rover moving.
3. Since we have no automated tests for perseus_input, a quick manual test matrix in the PR description (enable only / turbo only / both / neither, per controller type) would be good to have on record given this changes arming behaviour.

Inline comments for the smaller items.

GR3Y-SCALE:
Yep the change in deadman is desired, I talked with Dan and Matt about it as the current config is a little sus.
Basically the turbo button acts like a deadman, which is good for one handed control but to me that feels unsafe.

I will confirm with the rc link, since I hadn't had any issues and timeout worked correctly, but I will check.

PR #485, opened 2026-07-08
Isaac ROS NITROS/GXF from-source build for Orin + JetPack 7

DingoOz:
## Summary

NVIDIA's Isaac ROS binaries don't yet support Orin on JetPack 7 (Isaac ROS
4.x is Thor-only, 3.2 is JetPack-6-only). This branch is a staged, from-source
experiment to close that gap: build GXF/NITROS core and real AprilTag
detection directly against this hardware (L4T R39.2.0, CUDA 13.2,
Ubuntu 24.04), rather than waiting on NVIDIA's unported binaries.

- Stage 0-1: environment stand-up, confirmed NVIDIA's published
  `gxf_aarch64_cuda_13_0` (SBSA) binaries do load and initialize on real
  Orin/JetPack7 hardware — the key decision gate that avoided a much larger
  from-scratch GXF rebuild.
- Stage 2-3: full NITROS core build (with `$<INSTALL_PREFIX>` CMake
  generator-expression patches, git-lfs binary pulls, conda/native-glibc
  linking workarounds — all documented), minimal round-trip proof-of-life.
- Stage 4: real `isaac_ros_apriltag` detection, verified pixel-exact against
  NVIDIA's own ground-truth test fixture, then a live-camera pipeline
  (`pixi run test-apriltags`) using a Logitech C920, with a working RViz
  setup for remote viewing.
- Root-caused (not yet fixed) a `rectify`+`apriltag` crash on live frames as
  a genuine race condition, most likely in NITROS's cross-stream CUDA event
  synchronization for the non-NITROS-camera-to-`NitrosImage` type-adapter
  conversion path — very likely an upstream NVIDIA bug. Current live pipeline
  works around it by skipping `RectifyNode` (detection works; lens
  distortion uncorrected).

Nothing from NVIDIA's cloned repos is committed here (license terms
restrict redistribution) — only our own scripts, launch files, and docs
referencing them. Full runbook and findings at each stage:
`docs/source/systems/software/isaac-ros-nitros-source-build.md`.

This does not replace the existing JetPack-6-targeting Docker scaffold
(`software/docker/isaac-ros/`) — it's a separate, clearly-labeled
experiment track.

## Test plan

- [x] `colcon build` succeeds for the NITROS/apriltag workspace on-device
- [x] `apriltag_check.py` matches NVIDIA's ground-truth fixture pixel-exact
- [x] `pixi run test-apriltags` live-camera pipeline verified: detections
      publish at steady 10Hz, valid (non-NaN) pose on `/tf`, viewable in
      RViz from a remote laptop
- [ ] Real camera calibration for the physical C920 (still using borrowed
      placeholder intrinsics)
- [ ] Root-cause the rectify+apriltag race condition to an exact fix (or
      confirm/report upstream to NVIDIA)


DingoOz:
Opened in error against the wrong repo (this branch is from a hard-diverged fork, DingoOz/perseus-lite, not intended for upstream). Closing immediately.

